PR #1, opened 2015-03-13
Readme and source clarifications

anko:
No functional changes—just a bunch of proof-reading and formatting.


karlstav:
thanks, looks verry good!

PR #3, opened 2015-04-24
Fix initialisation segfault bug

anko:
Was dying with segfault on my machine. Effectively fixing up commit f0c1023.


karlstav:
my bad, thanks!

PR #8, opened 2015-04-25
More smoothing, looks better IMO.

CelestialWalrus:


CelestialWalrus:
Maybe we should add music genre presets.

    float smooth[64] = {1.5, 1.5, 2, 3, 3, 2, 2, 2, 1.5, 1.25, 1.25, 1.5, 1.25, 1.25, 1.5, 2, 2, 1.75, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.5, 1.75, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2};

^ this one's great for electronic music.

karlstav:
looks better indeed.  Presets also sounds like a cool idea. Although i have no idea of what your filter actually is doing. Could you give it a small technical description?

CelestialWalrus:
@karlstav some frequencies tend to be louder (e.g. bass) so this code just "equalizes" them.

karlstav:
@CelestialWalrus ok, there are already som constants (in the array k[]) that are trying to achieve just this. If it wasn't for those the higher frequencies whould be almost invisible compared to the lower. But as you point out this is diferent for diferent kinds of music. 

As my constants only work passivly, I can se that your filter activly works on the audio with some complex algorithm. Are you able to explain how it achieves its goal? I am verry curious to how it actually works (I know a little basic DSP, but i'm no expert.).

I think it would be usefull to stick a label on it also for further development.

CelestialWalrus:
@karlstav it's based on experimets I've conducted. It's not a complex algorithm. I just made it multiply the f[z] min(array)/array[freq].

PR #10, opened 2015-04-27
use canned recipe for compilation, add install/uninstall targets, turn on -Wall -Wextra

livibetter:
I decided to make this PR since it wouldn't involve the coding style of the C.

A few extra notes: I didn't include commonly use `-g -O2` since they were not here previously. You (@karlstav) can put it into `CFLAGS` if you want, so for `-std=c99` if you decide to go for that.

Two targets: `install` and `uninstall` are added for installation.

The only thing really change is `-Wall -Wextra`, lots of warnings. If you'd like me to help clean up those warnings once coding style (and C99) is decided and done..

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,24 @@ Getting started
 
     make
 
+### Installing
+
+Install `cava` to default `/usr/local`:
+
+    make install
+
+Or you can change `PREFIX`, for example:
+
+    make PREFIX=$HOME/.local install
+
+### Uninstalling
+
+    make uninstall
+
+Or:
+
+    make PREFIX=$HOME/.local uninstall
+
 
 Capturing audio
 ---------------
--- a/makefile
+++ b/makefile
@@ -1,6 +1,25 @@
-cava: cava.c
-	gcc cava.c -o cava -lasound -lm -lfftw3 -lpthread
-clean:
-	rm cava
+CC       = gcc
+CFLAGS   = -Wall -Wextra
+CPPFLAGS =
+LDFLAGS  = -lasound -lm -lfftw3 -lpthread
 
-.PHONY: clean
+INSTALL     = install
+INSTALL_BIN = $(INSTALL) -D -m 755
+
+PREFIX ?= /usr/local
+BINDIR  = $(DESTDIR)/$(PREFIX)/bin
+
+all: cava
+
+cava: cava.c
+
+install: all
+	$(INSTALL_BIN) cava $(BINDIR)/cava
+
+uninstall:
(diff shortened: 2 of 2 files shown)


karlstav:
@livibetter this look amazing, thanks. I'm a complete noob when it comes to makefiles. I guess the next step whit this whoudl be to make cava configurable so that one could compile it without the ALSA-part (mpd fifo only). Then one would not be dependent on the alsa_dev.

karlstav:
@livibetter I wasn't able to compile. 

seems that LDFLAGS places the -lm before the source file in the gcc command, so they where not used.

changing LDFLAGS to LDLIBS solved the problem.

I'm gonna change it in the makefile for now. If it's suposed to be LDFLAGS and I am doing something else wrong, please let me know.

livibetter:
Sorry, my bad, please move all of those to [`LDLIBS`](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html).

karlstav:
@livibetter I styled the code with astyle using the following options:

'astyle --style=linux --pad-oper --max-code-length=80 --keep-one-line-blocks --keep-one-line-statements --convert-tabs cava.c'

Does it look ok? I will try to keep the styling consistent from now on. 

Furthermore I'll go for C99, but my compiler fails on:

'cava.c:324:22: error: storage size of ‘action’ isn’t known
     struct sigaction action;'

but it compiles fine with -std=gnu99. Do you think its ok to use gnu99 instead? 

 If you'd help me clean up the warnings it would be great.

livibetter:
@karlstav I don't think it did much, for instance:

```
    if ((err = snd_pcm_open(&handle, device,  SND_PCM_STREAM_CAPTURE , 0) < 0))
        printf("error opening stream:    %s\n", snd_strerror(err) );
    else if(debug == 1) {
        printf("open stream succes\n");
    }
```

Notice the space between `if` and `(`.

It's best to do it by hand, I think I will handle that for you, clean up warnings, and also try to make it C99. Probably a day or two, is that okay for you?

karlstav:
@livibetter styled it again, now with:

'astyle --style=linux --pad-oper --pad-header --max-code-length=80 --keep-one-line-blocks --keep-one-line-statements --convert-tabs --indent=tab cava.c'

--pad-header should fix the spaces after if, while... and --indent=tab changes the indent to tabs instead of 4 spaces, which is actually more consistent with my coding style. @anko did the first auto styling. Before that it was complete anarchy. 

Just go ahead with additional styling cleanup, warnings cleanup and C99 (or gnu99?).

PR #14, opened 2015-04-29
make debug option

livibetter:
@karlstav could you clean up those codes in comments. I saw a lots of `debug` in comments, but I am not sure if I should just remove them.

If they are useful, utilize `DEBUG`; if not, it might be best to remove those, they make hard to read the codes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,16 @@ Getting started
 
     make
 
+You can use the following for compilation options, value in *italic style* is the default value:
+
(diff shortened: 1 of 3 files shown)


karlstav:
@livibetter I was just thinking about this. I'l merge your debug option, I think it's a very nice feature. Then I will clean up the comments. There are a lot of strange and confusing notes in there.

PR #20, opened 2015-05-04
add key controls

livibetter:
Using termios to set MIN=0 for not blocking, also handling `stty -echo`.


karlstav:
@livibetter can you show me how to merge this, there seems to be some conflicts because of the two previous merges.

livibetter:
@karlstav thanks to @anko merged for us. @karlstav it's now auto-merge-able.

PR #22, opened 2015-05-04
Merge livi's key controls patch

anko:
For #20. Was testing locally.

A bug I noticed is that if I start cava with `-S` for scientific mode and press the <kbd>s</kbd>-key to switch it to smoothed mode, the bars max out completely instead of showing smoothed bars. (This doesn't happen if you start in smoothed mode and use <kbd>s</kbd> to switch to scientific.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,3 +212,10 @@ Usage
 Exit with ctrl+c.
 
 If cava quits unexpectedly or is force killed, echo must be turned on manually with `stty -echo`.
+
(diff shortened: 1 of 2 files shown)


livibetter:
@anko good catch! It should be fixed now. Didn't know we have `<kbd/>` (or I forgot?), now using it, looking better.

anko:
Included in livi's PR now; closing.

PR #26, opened 2015-05-06
fix exit with no cleanup, change label "exit" to common "out", use EXIT_FAILURE

livibetter:
I swear I intentionally left a bug, so we could have a bugfix release the day after our first stable v0.1.0 release. This is all going according to my master scheme. (inb4 commit for major)

Just kidding, didn't test well.

The stream is failed to be opened, so cava `exit(1)` inside `music()`, and no cleanup. Hmm, well, doesn't sound like I create this bug, anyway, it should be fine now. I also make sure the error messages are printed to `stderr`.

The only thing is the message by the ALSA library is gone because of cleanup, but our error message is there, so we are still good, right?

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -76,10 +76,11 @@ music(void* data)
 //**init sound device***//
 
 	if ((err = snd_pcm_open(&handle, device, SND_PCM_STREAM_CAPTURE, 0) < 0)) {
+		cleanup();
 		fprintf(stderr,
 		        "error opening stream: %s\n",
 		        snd_strerror(err));
-		exit(1);
+		exit(EXIT_FAILURE);
 	}
 #ifdef DEBUG
 	else printf("open stream successful\n");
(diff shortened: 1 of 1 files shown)


karlstav:
so i guess this makes 0.1.1? I think this might be the reason why all the exit(1) were removed from the code in early early development.

Should I merge first then increment version?

CelestialWalrus:
@karlstav you should wait a few commits more before incrementing the version.

karlstav:
ok, so we will wait for a decent number of commits (if the changes aren't too big) until we release a new version (even though it's just a minor)?

CelestialWalrus:
@karlstav yeah, it's bad to change the number too often.

anko:
@CelestialWalrus Why is it bad? I tend to increment versions whenever I ① am confident that the changes that I've made are stable, and ② can be bothered. (Of course for 0.x.x versions it doesn't matter much.)

PR #30, opened 2015-05-17
General code improvements

sjef:
Commit messages should give a decent image of what has been changed. However, I'll list some of the main points.

* Declare increment variables inside the `for` loop control expression section. This is allowed in C99. This is probably a good idea for readability, as the reader can be sure the variable is only used in the loop. 

* Use ``while (true)`` instead of ``while (1)``. This is allowed in C99. Easier to read, compiles to the same code.

* Remove unused code. 

* Use correct std functions. A lot of the functions previously used expect variables of type `double` as arguments. We now use functions which expect `float` arguments.

* Use std function instead of macro.

* Use `static` for global variables.

* Make the variable `M` `const`.

* Use function prototypes.

* Improve the way `float` comparison is used.

* Add `f` suffix to make constants explicitly of type float.

* `Includes` have been sorted into a more visually appealing way.

We need a test suite.

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -72,7 +72,7 @@ void* input_alsa(void* data)
 	snd_pcm_uframes_t frames;
 	char *device = ((char*)data);
 	val = 44100;
-	int i, n, o, size, dir, err, lo;
+	int i, o, size, dir, err, lo;
 	int tempr, templ;
 	int radj, ladj;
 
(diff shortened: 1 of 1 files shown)


CelestialWalrus:
@sjef why have you closed it?

sjef:
@CelestialWalrus It's such a dramatic change. I'm no expert, this build should at least be tested by another person. 

I have tested it some more myself, and it does seem to work as it should. I'll reopen the PR.

sjef:
Oh, I see GitHub added a bunch more commits. Those should definitely not be merged yet, I'll close it (again) and fix my fork before attempting another PR.

Issue #4, opened 2015-04-24
Use ncurses instead of printing a lot?


anko:
At the moment, cava renders each frame by printing `term_width * term_height` characters to fill the terminal.

This could be sped up using [ncurses](http://en.wikipedia.org/wiki/Ncurses) to only erase and re-print what changed.

- - -

Example:

Going from bars like

    #   #
    ##  #
    ### #
    #####

to 

    #    
    ##  #
    #####
    #####

would just mean doing this (`d` for delete, `i` for insert)

        d
         
       i 
         

with two cursor jumps, instead of re-printing the whole lot.

karlstav:
Sounds like a good idea. I wasn't aware that such a thing as ncurses existed when i started working on cava. But reading about it now it sounds like just the thing. 

There is however a functionality in cava where it saves the last frame in a matrix and then only prints the changes. This saves a lot of CPU, but it still has to move the cursor along x number of characters where there is no changes. Ncurses might also improve the overall look, as it still is a bit glitchy in larger terminal screens. 

I'll note it down in the todo list! Hopefully I will get around to it soon, but it's hard to tell when. Feel free to have a crack at it. All bar heights are stored in the f[] array.

karlstav:
I started playing around with ncurses and i think the transition would be verry simple. Since ncurses prints to a buffer instead of printing directly to the stdio, we could just reprint each bar every cycle and then refresh.  The problem with printf is that it uses a lot of cpu, hence the complicated system of saving each frame in a matrix and then checking for changes every cycle. So ncurses would also save alot of cpu (cava still uses like 20-30% cpu on my rpi).


But I still haven't been able to print a single wide- (unicode) chrachter. And we are still going to need those in order to print the 1/8- 7/8 [block characters](http://unicode.org/charts/PDF/U2580.pdf) right? Or can anyone think of another trick to get the increased resolution?

I read that wide chrachters with ncurses in c programming could be a [problem](http://www.roguebasin.com/index.php?title=Ncursesw), help?

livibetter:
@karlstav that website seems to be blocking me for no reason.

Anyway, I have virtually zero knowledge of ncurses, but I do see [this](http://invisible-island.net/ncurses/ncurses.faq.html#no_line_drawing) a few times.

[...]

PR #33, opened 2015-05-20
Portable ncurses build

anko:
Got myself a Debian VPS for this…

I [previously](https://github.com/karlstav/cava/issues/4#issuecomment-102623327) didn't pass `--cflags` to `ncursesw-config`.
Turns out that's all that was needed.
It's quite underwhelming really.

Compiles on Arch, Debian 7 and Fedora. Fixes karlstav/cava#4.

@karlstav Could you check functionality? I only checked it compiles; ALSA loopback is a hassle.

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -21,7 +21,7 @@
 #include <time.h>
 #include <getopt.h>
 #include <pthread.h>
-#include "ncursesw/curses.h"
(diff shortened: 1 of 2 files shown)


karlstav:
works perfect! thanks.

PR #35, opened 2015-05-22
Merge ncurses-rendering to master

### LINKED ISSUE #4: Use ncurses instead of printing a lot?

anko:
Fixes #4. :sparkles: (Because merge-help [was requested](https://github.com/karlstav/cava/issues/4#issuecomment-103936472).)

I think this is right, but since I didn't write the feature, sanity-checks would be appreciated.

Probably warrants a minor-version bump to 0.2.0.


karlstav:
brilliant @anko ! bumbing up to 0.2.0

PR #37, opened 2015-06-01
Color Argument Changes

CZinsII:
I have added 'default' as a new value for the background color argument, which retains the default background color. This is useful for terminal emulators implementing transparency. Additionally, I modified the color argument handling to properly accept 'black.'

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -367,7 +367,7 @@ Options:\n\
 				}
 				break;
 			case 'c': // argument: foreground color
-				col = 0;
+				col = -2;
 				color = optarg;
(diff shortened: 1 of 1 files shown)


anko:
:+1:

I feel like this "default" would be the correct *actual default*—what you *should* get when you don't specify the `-C` flag. (It used to be, but got changed to black with the move from `printf` to `ncurses`. Maybe accidentally.)

karlstav:
great work @CZinsII ! 

@anko of course default should be default... Don't know why I made it force black as default. It might be because I didn't know about the use_default_colors function which I think we must use to retain the default background color.

PR #40, opened 2015-06-29
Added another mode & fixed the output method arg.

CelestialWalrus:


karlstav:
Looks very nice! Specially *the bass". 

I want to look through it more carefully before I merge. But I don't have time before next week. There has been a drop in activity the last couple of weeks and I'm afraid I will not have much time for this project until the end of the summer. But hold on, I have a lot of fixes and ideas I need realize! But no time now.

anko:
77d8133's bass-smoothing changes make the low end of the bars clip for a lot of my music. This may simply mean I have ridiculous music taste.

[superspink - *Wind*](http://bandcamp.superspink.org/track/wind) @ 1:00 :

bass clip

With `-m waves`, a lot of my stuff goes rectangular! :laughing:

BASS CLIP

Maybe we should have user-configurable smoothing parameters.

CelestialWalrus:
>Maybe we should have user-configurable smoothing parameters

@anko Any ideas for these? A config file, command-line argument?

CelestialWalrus:
This is the circle.

karlstav:
a lot of new things here..

I'll merge it and do some minor tweaking and bug fixing. 

Couple of suggestions (maybe i'll do it myself if I can find time):

1) Move the circle to a "mode" instead of a different output

2) Change the name of "mode" to "style"

I also really like the idea of a config file for default parameters (smoothing, input mode, output mode). I can see that it varies a lot with music style, terminal size and personal taste etc.. So it should be easy to play around with different variables and so forth. Maybe one also should be able to for example press 'g/G' to increase/decrease the gravity for falloff. Then pressing another key to save the parameters to the config file.

ahh so many ideas so little time, maybe I should quit my day job...

Issue #45, opened 2015-07-16
Target 'copyconf' fails if config-path does not exist.


phijor:
When trying to compile `cava` without having a path to the config-file, `cp` fails with the following message:

```
cp: cannot create regular file ‘/home/phijor/.config/cava/config’: No such file or directory
```

Steps to reproduce:

1. `rm -r ~/.config/cava` (that's where the config should up for me, because `$XDG_CONFIG_HOME` isn't set on my machine).
2. `make`.
3. See it fail.

I will try to fix the issue, shouldn't be to difficult. I think it's the best if `$CONFIGDIR` actually contains the directory the config will reside in, not the absolute path to the file itself. That way one can check if it actually exists before trying to put something in it.

karlstav:
thanks again!

PR #46, opened 2015-07-16
Create path to config-file if it does not exist.

phijor:
See issue #45.


karlstav:
ooops my bad. Thanks @phijor :wink:

PR #52, opened 2015-07-26
Properly add iniparser dependency

### LINKED ISSUE #49: Parallel make issue, e.g. with make -j5
```
# make -j5
[..]
gcc: error: ./iniparser/libiniparser.a: No such file or directory
[..]
```
The issue is missing dependencies. Target `cava` actually depends on presence of `iniparser/libiniparser.a` but the dependency is not coded into the Makefile.

anko:
Fixes #49.

As described in #49, compilation was failing when done in parallel.
This happened because the iniparser dependency wasn't expressed
explicitly as a make target dependency, but implicitly relied on timing.

### DIFF
--- a/makefile
+++ b/makefile
@@ -21,11 +21,11 @@ ifeq ($(debug),1)
 CPPFLAGS += -DDEBUG
 endif
 
-all: iniparser cava check-env copyconf
+all: cava check-env copyconf
 
-cava: cava.c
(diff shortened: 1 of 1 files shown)


karlstav:
great work @anko I'm no good at making make files :(

hartwork:
Nice!

PR #59, opened 2015-07-28
Honor system CFLAGS and CPPFLAGS

jubalh:
I think if we do it like that, the system can have control over the rest
of the flags.

### DIFF
--- a/makefile
+++ b/makefile
@@ -2,8 +2,8 @@ PACKAGE ?= cava
 VERSION ?= $(shell git describe --always --tags --dirty)
 
 CC       = gcc
-CFLAGS   = -std=c99 -Wall -Wextra
-CPPFLAGS = -DPACKAGE=\"$(PACKAGE)\" -DVERSION=\"$(VERSION)\" \
+CFLAGS   += -std=c99 -Wall -Wextra
+CPPFLAGS += -DPACKAGE=\"$(PACKAGE)\" -DVERSION=\"$(VERSION)\" \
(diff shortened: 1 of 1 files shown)


jubalh:
I think like this it's better. What do you think? Am I missing something here?

karlstav:
I have no idea, I don't know anything about makefiles. The guy who made the original make file @livibetter f7c00238c5e1e011b4245308c1eb66ef2f35c7ea is no longer active in the project. Any second opinion will be much appreciated. But if you feel confident about this, I'll just merge it.

livibetter:
It's fine, just the align is a bit off.

anko:
+1 from me too.

Brief explanation for @karlstav & others interested: `CFLAGS` is one of *make*'s [default variables](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html#Implicit-Variables), which each have a specific purpose—this one's for passing flags to the C compiler. Some systems set default values for it. For example, [Arch in `/etc/makepkg.conf`](https://wiki.archlinux.org/index.php/Makepkg#Configuration) and [Gentoo in `/etc/portage/make.conf`](https://wiki.gentoo.org/wiki/CFLAGS). Setting a variable using `=` changes its contents, while using `+=` appends to them. Here that ensures the system-defined compiler flags are kept too.

Example use-case: system default `CFLAGS` telling the compiler to optimise specifically for the current processor's architecture using [`-march=native`](https://gcc.gnu.org/onlinedocs/gcc-4.9.3/gcc/i386-and-x86-64-Options.html).

karlstav:
sounds great, thanks!

Issue #55, opened 2015-07-27
Idea: Check for /sys/module/snd_aloop/ if input is ALSA


hartwork:
The error "not permitted"
```
ALSA lib /var/tmp/portage/media-libs/alsa-lib-1.0.29/work/alsa-lib-1.0.29/src/pcm/pcm_hw.c:1693:(_snd_pcm_hw_open) Invalid value for card
error opening stream: Operation not permitted
```
is a bit misleading, since adding `sudo` will not fix the problem. It would be cool  if cava could check for the existence of /sys/module/ (sysfs mounted) and then for /sys/module/snd_aloop/ (module loaded). That should allow producing a more helpful error message. What do you think?

karlstav:
agreed, getting the alsa/pulseaudio stuff to work can be difficult and not all people like reading readme files...

I can get to work on it next week, I hope.

Any suggestions would be awesome.

hartwork:
Cool!

karlstav:
guess this one is done also, thanks to @hartwork :+1:

PR #63, opened 2015-08-08
With ALSA input, check if the snd_aloop module is loaded

hartwork:
Hi! Here's a patch for issue #55. Please consider application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ Usage
         -b 1..(console columns/2-1) or 200  number of bars in the spectrum (default 25 + fills up the console), program will automatically adjust if there are too many frequency bands)
         -i 'input method'           method used for listnening to audio, supports: 'alsa' and 'fifo'
         -o 'output method'          method used for outputting processed data, supports: 'ncurses', 'noncurses' and 'circle'
-        -d 'alsa device'            name of alsa capture device (default 'hw:1,1')
+        -d 'alsa device'            name of alsa capture device (default 'hw:Loopback,1')
         -p 'fifo path'              path to fifo (default '/tmp/mpd.fifo')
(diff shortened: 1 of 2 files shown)


karlstav:
I'm not sure, but people might try to use cava with another type of input like a microphone. (I do it sometime does to show it off....) .

However I think this check could be very useful in the "error opening stream: " error in alsa.c. There it could check if the user hasn't loaded the loopback or is just trying to open the wrong device.

hartwork:
The patch as of now only does that check for ALSA input mode.
Are you referring (a) to non-ALSA input or (b) to ALSA input with something other than a loop device?
Case (a) is covered already.
For case (b) we could turn the error into a warning or add a command line option to skip the check. I like former better as of now.

Best, Sebastian

karlstav:
I was referring to b, other alsa input than loop device.

If we change cava from using 'hw:1,1' as default to 'hw:Loopback,1', we could skip your check whenever the user specifies another device.

hartwork:
I have implemented and pushed those ideas now.

These are the places where I wasn't sure if hw:0/1 should have been touched, too:
```
# grep -R 'hw:[0-9]' .
./example_files/etc/asound.conf:            a { channels 2 pcm "hw:0,0" }  # the real device
./example_files/etc/asound.conf:            b { channels 2 pcm "hw:1,0" }  # the loopback driver
./example_files/config:; source = hw:1,1 # ALSA device or FIFO path.
```

hartwork:
PS: I wonder if `Loopback` is case sensitive or not.

karlstav:
the config file and example asound.conf can be changed to, I'll take care of it. 

Yes it is case sensitive, capital L, so everything looks good.

hartwork:
Excellent!

PR #64, opened 2015-08-12
MAKEFILE shouldn't copy any files into user's ~

CelestialWalrus:
Namarrgon | CW: the buildsystem is copying files into my ~      

Also, @karlstav, could you release 0.3.3? Thanks.


CelestialWalrus:
Disregard this. (but you should release 0.3.3 now, 'cause of all the changes ;p)

Issue #68, opened 2015-09-06
Numbers being displayed instead of bars


aouelete:
Here's a [screenshot](https://i.imgur.com/MdX2QVe.png) of what I am seeing. If I run cava in the tty, the bars are displayed correctly, so this makes me think that it is a font issue, but I have tried different fonts with nothing changing.

anko:
I'd also guess a font issue.

cava ships with a font `cava.psf` which it sets on the TTY to ensure it can draw bar characters. That doesn't apply to terminal emulators.

What terminal emulator are you using? What fonts have you tried?

- - -

Latest commit 9712160d1d82ee0c273ae364b578a171e1982137 on `master` works as expected for me, on [st](http://st.suckless.org/), [evilvte](http://www.calno.com/evilvte/) and [rxvt-unicode](http://software.schmorp.de/pkg/rxvt-unicode.html) and [gnome-terminal](https://help.gnome.org/users/gnome-terminal/stable/) with a few monospace fonts I tried.

karlstav:
If this is what you are seein in terminal emulator, then the problem is that cava thinks it is in the tty. 

As anko said cava uses its own font in tty, cava uses the return value from trying to set the font in order to determine wether it is in tty or not: 

'virt = system("setfont cava.psf  >/dev/null 2>&1");'

This returns 0 in tty and some other value in virtual terminal, but maybe there are some better way of ensuring whether running in tty or virtual terminal.

aouelete:
I just tried Droid Sans Mono, Fira Mono, DejaVu Sans Mono, and Liberation Sans Mono in urxvt, problem is still present. Also, cava used to work for me, I don't know exactly when the bars changed to numbers, but it seems to be a more recent issue.

anko:
@karlstav [Perhaps](http://stackoverflow.com/questions/2908044/how-to-tell-if-running-in-a-linux-console-versus-an-ssh-session) we should use [`ttyname`](http://man7.org/linux/man-pages/man3/ttyname.3.html).

Quick test—

```c
#include <unistd.h>
#include <stdio.h>

int main() {
    printf("%s\n", ttyname(0)); // Check where stdin is coming from
}
```

—returns `/dev/pts/19` in a terminal emulator and `/dev/tty2` in TTY 2.

I'll test a little more and send a patch.

[...]

PR #69, opened 2015-09-06
Use ttyname to check whether in a TTY

anko:
More robust, as it gets nearer the source instead of relying on the exit code of a shell command.

Might help with #68. @aouelete, could you check with your setup?


aouelete:
It looks like this fixes the isssue, I don't see any problems. Thanks!

karlstav:
Great work @anko!

PR #72, opened 2015-10-01
Fix M_PI redefinition warning

hesiod:
Should be uncontroversial.


karlstav:
looks good. 

@hesiod, this warning suddenly started popping up, any idea why? Did they suddenly include a definition of M_PI in math.h or something?

PR #80, opened 2015-10-31
Added error handling for when config cant be accessed

binarii:
I had initially installed as sudo, causing the default configuration folder to be owned by root. When running Cava, this resulted in a segmentation fault. This can be fixed by modifying ownership easily, however it was not readily apparent this was the cause. I added some error handling to when the configuration is opened to make it more apparent for users.


karlstav:
nice thanks!

karlstav:
@binarii btw, I see you used 'print', I changed it to 'printf', I hope thats what you meant to write.

PR #90, opened 2016-02-19
Fix typo then→than

sinisterstuf:


karlstav:
Thanks for the spellcheck, I need it, there are probably a billion more typos hidden away in here.

PR #98, opened 2016-04-12
Reload config file on SIGUSR1

EmilGedda:
Cava now also listens to SIGUSR1, and upon recieving USR1 cava will reload its config, just like pressing 'r'. Before there were no easy way of reloading the config from a script.
Just a minor patch, making use of the existing signal handler and config-reloading trigger.

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -86,6 +86,8 @@ int stereo = -1;
 int M = 2048;
 char supportedInput[255] = "'fifo'";
 
+// whether we should reload the config or not
+int should_reload = 0;
 
 
 // general: cleanup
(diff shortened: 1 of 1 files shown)


cava.c
```diff
@@ -628,15 +635,13 @@ as of 0.4.0 all options are specified in config file, see in '/home/username/.co
 	#endif
 
 	if (im == 2) {
-		thr_id = pthread_create(&p_thread, NULL, input_fifo,
-														(void*)&audio); //starting fifomusic listener
+		thr_id = pthread_create(&p_thread, NULL, input_fifo, (void*)&audio); //starting fifomusic listener
```
EmilGedda: Not related to the SIGUSR1 implementation, just removal of unnecessary spacing

karlstav:
looks brillliant. 

@EmilGedda could it be an idea to put something in the readme about how to use ths?

EmilGedda:
Sure, I'll get right on it.

PR #103, opened 2016-06-02
ignore all the files created during the compilation

kowalcj0:


HaCk3Dq:
omg, why do u dot just ignore folders?
but ignoring every single file -_-

anko:
@HaCk3Dq 

> why not just ignore folders

Assuming you mean `iniparser/`: some files in there are intentionally checked in to the repository, like `iniparser/src/iniparser.c`. An ignore rule for the whole folder would hit those too.

> you should never ignore Makefile

The `Makefile` is generated by [automake](https://en.wikipedia.org/wiki/Automake) when running `./autogen.sh` and `./configure`. The one that _really_ shouldn't be ignored is `Makefile.am`, as correctly excluded here.

karlstav:
@kowalcj0 can I pull this? Is it somehow related to your other PR?

kowalcj0:
Hi @karlstav yes it's related to the other PR. Sorry for such delayed answer but I'm bogged down with work chores atm. Will try to sort out the other PR in next few days.

PR #111, opened 2016-09-16
add stdint header for compilation in Raspbian

VectorCell:
In Raspbian (and perhaps in other systems) there is a compilation error if stdint.h is not included in output/raw.c, because otherwise the int types are not defined (uint8_t and uint16_t).


karlstav:
Oops. Thnaks a lot! I should really get some kind of automatic testing implemented.

Issue #91, opened 2016-03-05
More color support (256/rrggbb/true colors)


MCManuelLP:
Until now the only way to change the available colors is to change the default terminal colors, is it possible to support more colors out of the box in the future?

karlstav:
@MCManuelLP available colors will be limited to the colors supported in the terminal emulator. I guess 256 colors is possible like in [cli-visualizer](https://github.com/dpayne/cli-visualizer#256-colors). Looking at [this](https://gist.github.com/XVilka/8346728) I guess more colors are also possible. There is no specific plan yet, but I'll put it down in the todo list.

hashhar:
I think you should add a configuration option called 'truecolor' which can be an integer from 0 to 255 which will be simply passed to `ncurses`. Currently you are reading from the config file and using a string compare and assign values to `bcolor` and `color` accordingly.

I propose the following:
1. Add a new option called `BG Truecolor` and `FG Truecolor` which will be an `int` (a `char` really) from 0 to 255.
2. Ignore the `bcolor` and `color` if either of the above are set and simply set `bcolor=BG Truecolor` and `color=FG Truecolor`.
3. Marvel at rainbows!!! :tada: 

If you want I can submit a pull within the next few days after properly testing that nothing else breaks because of such a change.

**PS: Thanks a lot for such a great tool @karlstav.**

karlstav:
@hashhar yes indeed please do!

But I suggest a simple "custom" instead of "BG Truecolor"/"FG Truecolor". and then you whould need six options: red, green and blue for both foreground and background. Which must be an int from 0-1000 each.  Why 1000? because that's how the [color redefinition](http://tldp.org/HOWTO/NCURSES-Programming-HOWTO/color.html#CHANGECOLORDEFS) in ncurses works. Off course we don't have to use 0-1000, but I don't see why not.

Note: we need to make a handle for terminals who can't change the color definition.

hashhar:
@karlstav Sure I'll start working on it.

hashhar:
@karlstav Hey, I managed to add a config setting but there's an issue I would like to discuss with you.

**NOTE: I'm currently only looking to get the feature going as a proof of concept and will then streamline the config and add tests.**

I am currently designing the patch as follows.

1. A new config. `customColor` and `customBColor`. Both have to be valid html hex colour codes.
2. Read it in and validate the values. Exit after errors if the values are not valid.
[...]

PR #115, opened 2016-10-27
Add support for more colors in ncurses output. WIP for #91

hashhar:
WIP for #91 

**NOTE: This breaks the `b` and `c` keys for cycling through the colors
in ncurses mode.**

Added a new function in cava.c `int validate_color(char
*colourToValidate, int outputMode)` which returns 1 if the colour is
valid, 0 otherwise.

Validation is done to ensure that HTML colours are of the form `#xxxxxx` or a
named ncurses colour is present.

HTML colours are allowed only when output mode is ncurses (ie `om = 1 ||
om = 2`). HTML and named colours can be mixed.

If the output mode is other than ncurses, the old method of setting
`bgcol` and `col` will be used.

Added a function in `terminal_ncurses.c` called `struct cols
parse_color(char *col)` which returns a struct containing:
1. `col`: The ncurses colour number parsed.
2. `R`, `G` and `B`: Integers that represent the R, G and B values
   parsed.

The `init_terminal_ncurses()` method now accepts two character pointers
instead of integers. The pointers contain color names (either HTML or
ncurses names) which are then parsed by the parse function defined
above.

Currently there is no check to find if the terminal supports color
redefinition or not. According to the parsed value we decide whether to
perform redefinition or use the predefined ncurses colours.

Ncurses colours 1 and 2 are redefined for foreground and background
respectively. This means that upon exiting, all the occurrences of 1 and
2 would have been redefined. To fix the colours back to the original the
user has to open another terminal.

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -23,6 +23,7 @@
 #include <getopt.h>
 #include <pthread.h>
 #include <dirent.h>
+#include <ctype.h>
 
 
 #ifdef NCURSES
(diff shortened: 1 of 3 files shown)


hashhar:
TODO:

- [ ] Add a check for deciding if color redefinition is possible or not.
- [ ] Fix the broken `b` and `c` key functionality.
- [ ] Write tests and modularize the code to make it easier to read and reduce chances of regressions.

hashhar:
I'm currently upgrading my system so I may not be able to do any work for the day. 😞

karlstav:
@hashhar I finally got around to check this out. Looks really really good! Also it looks like it's actually 24-bit? Since ncurses is not using the colors already defined by the terminal, but redefining new colors instead.

Do you think I should merge it yet? Or maybe create a separate branch for it?

hashhar:
I made some changes. I have yet to push them. I will try to push them tomorrow.

I added code to check if the terminal supports color redefinition or not and simplified the code a little bit.

But I was not able to fix the broken `b` and `c` keys feature. I think you should create a separate branch for the time being while I try to fix those issues.

karlstav:
created new branch `more-color`, I don't know if I can "move" this PR or if we have to close this and you create a new PR to the new branch?

hashhar:
Hey, I think is good enough for now. Maybe you should merge this and we can see if it has any issues for people who want to try it.

The only things that will not work are:

1. `b` and `c` keys to cycle between colors. I am trying to fix it but time is an issue.
2. No errors are currently displayed if the terminal doesn't support color redefiniton, I simply fallback to the default colours in that case.

karlstav:
some thoughts (if interested):

 1. maybe just disable color cycling when a custom color is defined? (I don't think it will be missed)
 2. an error and exit would be nice, so people don't just get mad over it not working

I might look into it over the weekend

hashhar:
Thanks. I'll add the error and exit and I think I'll disable the color cycling for now.

PR #120, opened 2016-12-03
New update/features for the GUI branch.

nikp123:
This is a update to the gui branch, that was discussed about earlier.

Changelog:
*SDL2 mode is now fixed (colors and reloading now work)
*Fixed a few memory leaks
*Updated config and README to suit recent changes
*Updated configure.ac so it can compile SDL2 mode
*Added titlebar name for X mode
P.S. : Apologies for recent inactivity, had been busy in school :P


nikp123:
*Also added changing bar spacing with 'a' and 's' keys.

PR #124, opened 2016-12-29
Data range check for binary output

worron:
Faced with the problem of wrong peaks for binary output. Not sure this patch is perfect, but seems it works.


karlstav:
@worron looks very good, seems i forgot a check for that

PR #134, opened 2017-02-17
terminal_ncurses: refactoring

chelovechishko:


output/terminal_ncurses.c
```diff
@@ -1,145 +1,142 @@
...
-				if (tty) for (q = 0; q < bw; q++) mvprintw((h - f[i] / 8), i*bw + q + i*bs + rest, "%d",(f[i] % 8) );
-				else for (q = 0; q < bw; q++) mvaddwstr((h - f[i] / 8), i*bw + q + i*bs + rest, bh[(f[i] % 8) - 1]);
+			if (f[bar] % 8) {
+				if (is_tty) {
+					for (int width = 0; width < bar_width; width++)
+						mvprintw((height - n), CURRENT_COLUMN, "%d",
+								(f[bar] % 8));
+				} else {
+					for (int width = 0; width < bar_width; width++)
+						mvaddwstr((height - n), CURRENT_COLUMN,
```
chelovechishko: Here is 'n' which is not inside its loop. It supposed to be 'f[bar] / 8' or what?

karlstav: yeah the last loop with 'n' runs to 'f[bar] / 8', so I guess you can/should use that instead.

If it helps you anything this is where it draws the bar upwards from the last position (it only draws changes from the last 'frame'), in addition to drawing block characters cava draws 1/8-7/8 fraction block characters on the top of the bar to increase resolution, which is why 'f' is multiplied by 8. This particular 'mvaddwstr' is the drawing of the top bar, the height of it is of course where the last loop ended 'n', which is of course also 'f[bar] / 8'. Maybe you got this all figured out already.

karlstav:
broken :( it just fills terminal with foreground color.

chelovechishko:
How did you check it? For me gnome-terminal works as expected. xterm don't change background color as before, but works too.

karlstav:
It only happens when defining foreground hex color and leaving background as default. Also the bar width or spacing is changed.

chelovechishko:
Didn't see bar width / spacing changes. Is it alright now?

karlstav:
looks good now. 

again, thanks a million!

Issue #84, opened 2015-12-12
Error on Raspbian at run-time.


0xNear:
```bash
pi@raspberrypi:~/Desktop/cava $ cava
cava: error while loading shared libraries: libiniparser.so.4: cannot open shared object file: No such file or directory
```
It's not running under Raspberry Pi 2B. I have installed Raspbian GNU/Linux 8.0 (jessie).
There was no errors when I was compiling, I compiled it two times for being sure.

karlstav:
@0xNear I think I might have stumbled upon the bug that you are experiencing. I can see that you have installed cava to your system. For some reason when running from system installation on rasbian (I have version 7 ) it can't find `libiniparser.so.4`. It works if you run it from the build tree (./cava). 

`libiniparser.so.4` is located in the exact same place as on my ubuntu system, in `/usr/local/lib`, which is where  iniparser says it should be. On ubuntu everything works fine as well, so I'm lost here...

PlasmaBinturong:
Hey, I encounter the same error on Ubuntu 14.04. I managed to install it however, by doing:

    export LD_LIBRARY_PATH=/usr/local/lib

before the compilation steps.
Not sure that is the proper way to do it but I know nothing about C++, compilation, linker flags...!

interestingly, it also works if I do `./configure --prefix=/home/me/.local`.

I can show you the output I got before fixing, if you need it.

Thanks for this program, it looks great! =D

PS: for an enhancement, would it be possible to set the background color to transparent? Would be great in a xterm =)

karlstav:
@PlasmaBinturong that looks programming. I think this can be fixed easily in the configure-script/makefile. This issue is getting old, I should have gotten around to fixing it sooner, but have been lazy...

As for transparent background I, think the default is supposed to be "leave untouched". So if you leave it commented in the config and don't touch 'b' while running it's supposed to stay transparent if it where. The background transparency is a feature of the terminal emulator, not the application running in it. So cava itself cannot make the background transparent, if it isn't already. 

I can see now that the config file says default background color is black. That should say "leave untouched".

PlasmaBinturong:
I installed it correctly now, the thing to do was `export LD_RUN_PATH=/usr/local/lib` before compilation.

This message during `make install` helped me:

    Libraries have been installed in:
       /usr/local/lib

    If you ever happen to want to link against installed libraries
    in a given directory, LIBDIR, you must either use libtool, and
    specify the full pathname of the library, or use the `-LLIBDIR'
[...]

PR #144, opened 2017-03-22
Remove cava_LDFLAGS

jubalh:
Not needed.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -11,7 +11,6 @@ M_CPPFLAGS = -DSYSTEM_LIBINIPARSER=@SYSTEM_LIBINIPARSER@
 bin_PROGRAMS = cava
 cava_SOURCES = cava.c
 cava_LDADD = -liniparser
-cava_LDFLAGS = -L/usr/local/lib -Wl,-rpath /usr/local/lib
 cava_CPPFLAGS = -DPACKAGE=\"$(PACKAGE)\" -DVERSION=\"$(VERSION)\" \
            -D_POSIX_SOURCE -D _POSIX_C_SOURCE=200809L
 cava_CFLAGS = -std=c99 -Wall -Wextra -Wno-unused-result -Wno-maybe-uninitialized 


karlstav:
are you sure? I must have added them there for a reason, or maybe I just copied them from some example file. What would these flags normally be for?

jubalh:
/usr/local/lib should automatically be available (/etc/ld.so.conf should check for /usr/local/lib64 and /usr/local/lib, /usr/lib and /usr/lib64 are checked already).

there is no reason to use [rpath](https://en.wikipedia.org/wiki/Rpath) is not needed either as those directories are already checked so dont need a preference.

karlstav:
seems reasonable

karlstav:
@jubalh did you consider #84 when you did this?

jubalh:
I dont see a problem.

anned20:
It broke it...

karlstav:
@anned20 have you looked in to #84?  Is it the same problem?

anned20:
Yeah it is. I tried to do that and it worked instantly.

jubalh:
#84 is a hack, not a good solution.

@anned20 which distro?

anned20:
Linux Mint 18.1 Sarah

karlstav:
I am also on mint 18.1, I also get this message during `make install`:

```Libraries have been installed in:
   /usr/local/lib

If you ever happen to want to link against installed libraries
in a given directory, LIBDIR, you must either use libtool, and
specify the full pathname of the library, or use the `-LLIBDIR'
flag during linking and do at least one of the following:
   - add LIBDIR to the `LD_LIBRARY_PATH' environment variable
     during execution
   - add LIBDIR to the `LD_RUN_PATH' environment variable
     during linking
   - use the `-Wl,-rpath -Wl,LIBDIR' linker flag
   - have your system administrator add LIBDIR to `/etc/ld.so.conf'

See any operating system documentation about shared libraries for
more information, such as the ld(1) and ld.so(8) manual pages.
```

But it works. 

What do you say @jubalh? Restore the LDFLAGS? I agree that it doesn't seem necessary, but  it does solve the problem. I personally do not know enough abut these things to understand what the problem is or what causes it or how to solve it properly. :frowning_face:

ibrokemypie:
How do you change the value/ignore that flag while configuring? No longer builds on arch linux since they removed /usr/local, though builds fine when changed to just /usr

jubalh:
> No longer builds on arch linux since they removed /usr/local, though builds fine when changed to just /usr

@ibrokemypie Arch Linux removed `/usr/local` ?

> How do you change the value/ignore that flag while configuring?

```
export LDFLAGS=-L/usr/local/whatever
./configure # run configure make or whatever
```

Alternatively you can just define it before the command.

PR #148, opened 2017-03-28
Remove test line

jubalh:
I dont see a reason for this. I suppose it was used for testing.


karlstav:
probably, I wasn't the one who added it

Issue #147, opened 2017-03-28
Installing font globally


jubalh:
In https://github.com/karlstav/cava/blob/master/cava.c#L697 to load the cava font.

In https://github.com/karlstav/cava/blob/master/Makefile.am#L18 you define the directory for the font as `/usr/share/consolefonts`.

My `man setfont` sais:
```
/usr/share/kbd/consolefonts  is the default font directory.  /usr/share/kbd/unimaps is the default directory for Unicode maps.  /usr/share/kbd/consoletrans is the default directory for screen mappings.  The default font is a file default (or default8xN if the -N option was given for some number N) perhaps with suitable extension (like .psf).
```

- [ ] Use /usr/share/kdb/consolefonts as directory
- [ ] Actually install the font and not just define in the configure script

Honestly I am not sure if my directory is an openSUSE specific thing. In this case it should be made variable.

Issue #116, opened 2016-10-29
8's instead of bars, cava.psf not loading


amtino:
Similar to Issue #68, I am getting numerals 1-8 in Cava instead of bars, when run from TTY. The bars appear fine from my Raspberry Pi 3's terminal emulator, and via SSH on my MacBook Pro's Terminal. Running latest version of Raspbian. I have cava.psf in my /cava directory but also copied it to ./config/cava for safe measure but still no luck in TTY.

amtino:
Solved: moving cava.psf to user directory (home/pi) makes cava display bars in TTY as expected

karlstav:
@amtino I realize now that this is actually a bug. It will work as long as cava is executed from the same path as cava.psf is located in. The problem is that installing cava to the system `sudo make install cava` should also copy cava.psf to some system font directory. But I have forgotten to implement it.

I'll reopen this, so that I will remember to fix it when I have time.

karlstav:
@amtino if it's still interesting I fount out that copying `cava.psf` to: `/usr/share/consolefonts/` is the actual solution for this issue. At least for debian-based systems. However that path might be different in other distros so I need some kind of distro check to add it to the makefile.

PR #150, opened 2017-03-28
Add font directory environment variable

jubalh:
Adds a variable to specify the location where the font will be installed.

In this patch I chose to use /usr/share/consolefonts as the default,
though on my openSUSE system it will be /usr/share/kbd/consolefonts.
Maybe it varies between distros, so its good to have a variable.

Example usage:
FONT_DIR="/usr/share/kbd/consolefonts" ./configure && make && make install

Fixes https://github.com/karlstav/cava/issues/147


karlstav:
this also relates to #116. this path do vary between distroes, what would be awesome was something that would check the distro/find the consolefont dir automatically.

PR #152, opened 2017-04-09
Some changes to the code

nikp123:
Changelog: 

- Migrated code from cava.c to output (removes code clutter)
- Fixed SIESEGV when changing color
- All window options are migrated to a specific category in the config file
- Synced the fork (obviously)
- Updated the documentation to follow the given changes

I was wondering, is it ready for merging with the main branch? Or does it need more work :\


karlstav:
I'll have to look more closer through it first. But I still think this might be better off as a separate branch.,

PR #154, opened 2017-04-16
raw: refactoring

chelovechishko:
Didn't check it.

### DIFF
--- a/output/raw.c
+++ b/output/raw.c
@@ -1,60 +1,48 @@
 #include <stdio.h>
-#include <stdlib.h>
 #include <stdint.h>
 
+#define TEN_THSND 10000
+#define BIT_16 16
+#define BIT_8 8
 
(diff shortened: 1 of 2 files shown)


karlstav:
ascii still works, but I'm not getting any output from binary :frowning_face:

karlstav:
OK, now I got something. I don't have an application to test if it is correct. But then again I never had. Merge?

chelovechishko:
Sorry @worron Could you check binary output, if you use it?

worron:
Just tried 16bit and it works all ok. I can check 8bit if it necessary, but tomorrow only, it's need a bit time to edit my scripts for 8bit output.

chelovechishko:
Not necessary, the code for both parts is the same. Thank you @worron

karlstav:
I'll merge it then. Thanks again @chelovechishko for the refactoring.

PR #183, opened 2017-10-29
Add sndio backend

t6:
This adds an [sndio](http://www.sndio.org/) backend to cava.

sndio is the audio framework used on OpenBSD, but it's also available on
FreeBSD and Linux.  So far I only tested it on FreeBSD.

To test it
```bash
# Start sndiod with a monitor sub-device
$ sndiod -dd -s default -m mon -s monitor

# Set the AUDIODEVICE environment variable to override the default
# sndio device and run cava
$ AUDIODEVICE=snd/0.monitor cava
```

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -1,5 +1,5 @@
 #define _XOPEN_SOURCE_EXTENDED
-#include <alloca.h>
+#include <stdlib.h>
 #include <locale.h>
 #include <stdio.h>
 #include <stddef.h>
@@ -51,6 +51,10 @@
 #ifdef PULSE
 #include "input/pulse.h"
 #include "input/pulse.c"
+#endif
+
+#ifdef SNDIO
+#include "input/sndio.c"
 #endif
 
 #include <iniparser.h>
@@ -287,6 +291,9 @@ as of 0.4.0 all options are specified in config file, see in '/home/username/.co
     #ifdef PULSE
         strcat(supportedInput,", 'pulse'");
     #endif
+    #ifdef SNDIO
+        strcat(supportedInput,", 'sndio'");
+    #endif
 
 	//fft: planning to rock
 	fftw_complex outl[M / 2 + 1][2];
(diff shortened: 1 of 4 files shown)


karlstav:
Hi @t6 
I'm not familiar with sndio or BSD in general, but great that you wish to expand cava in to that realm.

however it seems you missed a `ifdef SNDIO`:

```
In file included from cava.c:63:0:
config.c: In function ‘load_config’:
config.c:418:69: error: ‘SIO_DEVANY’ undeclared (first use in this function)
  p->audio_source = (char *)iniparser_getstring(ini, "input:source", SIO_DEVANY)
                                                                     ^
config.c:418:69: note: each undeclared identifier is reported only once for each function it appears in
Makefile:528: recipe for target 'cava-cava.o' failed
```

t6:
> however it seems you missed a ifdef SNDIO
Woops, fixed it. Thanks!

karlstav:
ok, it compiles now.

If you want you can write it into the readme under 'capturing audio', just add another section.

karlstav:
there is a problem compiling on rasbpian:

```
cava-cava.o: In function `initialize_audio_parameters':
/home/pi/cava/input/alsa.c:20: undefined reference to `alloca'
collect2: error: ld returned 1 exit status
Makefile:483: recipe for target 'cava' failed
make[1]: *** [cava] Error 1
```

karlstav:
I can see that you removed `alloca.h` any reason why?

Also I can see that you added `stdlib.h`, probably a reason why I didn't include it in the first place, but I can't remember.

t6:
Thanks for merging!

> I can see that you removed alloca.h any reason why?

On FreeBSD alloca() is defined in `stdlib.h` and we don't have `alloca.h`.  I did this blindly to make cava compile.  I thought this was a safe change to make but apparently it's problematic.  Sorry about that.

I see now that the only reference to alloca() is in `input/alsa.c`, so I think the best way to solve it is to just move `#include <alloca.h>` to `input/alsa.c`.  I doubt the ALSA input is even going to work on FreeBSD.  I can prepare a PR for it if you want.

karlstav:
I made a fix by not including stdlib except when sndio is defined and to include alloca when sndio is not defined. It's a bit ugly, but as long as no one has installed both sndio and alsa on a raspbian I thinks we will be fine.

t6:
I guess that'll do.  Though I think with that solution there might be problems when users try to build cava on FreeBSD with PulseAudio support but don't have sndio installed.

karlstav:
could you make a detection of OS in the configure script, so there will be a `ifdef BSD` and `ifdef LINUX`?

t6:
We can probably add a check for `alloca.h` instead using `AC_CHECK_HEADER`. Then we can use `HAVE_ALLOCA_H` or something similar.

PR #190, opened 2017-12-09
[feature] Clock feature

syfluqs:
A simple clock overlay that can be toggled on/off with <kbd>=</kbd>. The positions can be changed/cycled-through with <kbd>+</kbd>.

The ini key "general:show_clock=1" will enable the clock and the initial position of the clock can be set with "general:clock_position" key.


karlstav:
closing off as not relevant.

PR #208, opened 2018-04-29
Document keys in help text

winny-:
I didn't find the keys documented outside of the readme, figured this is a nice place to put them :)


karlstav:
yea why not.

Issue #211, opened 2018-05-02
fftw_complex type is misused


ginkage:
So, there is a nasty bug. The fftw_complex type is defined as double[2], yet cava is treating it as a single double, leading to, basically, only using real parts of the output (e.g., if fftw produces an output of a+ib, c+id, e+if, g+ih, cava will interpret it as a+ic, e+ig). All imaginary part, and all of the second half of the output is ignored.
I'm not sure if that's the reason for the comment "maybe the nq freq is in M/4"... It sure seems to be related.

karlstav:
this is a nasty bug indeed, and must be related to the mystery of the nyquist frequencies being in M/4 and not M/2 where it is supposed to be.

But I need some more explaining.

what is the problem when I'm using `fftw_complex outl[M / 2 + 1][2]`, instead of `fftw_complex outl[M / 2 + 1]`?

ginkage:
The problem is that, instead of double[M/2+1][2], you have outl/outr declared as double[M/2+1][2][2], so the resulting array is twice the size it needs to be.
Fftw then happily fills the first half of that array, and then, when you dereference it like *out[i][0] and *out[i][1], you access elements out[i][0][0] and out[i][1][0] respectively, which is basically every second output element, which is why I'm saying that you ignore the imaginary parts.

karlstav:
aha I get it now! 

As you say line 578 must now be changed to M /2. 

I gave it a quick test and it looks much more precise then it did did before! Thanks a million for spotting this bug.

PR #212, opened 2018-05-02
Fix fftw_complex usage

ginkage:
See issue #211 for the back story.


ginkage:
Not sure how to edit pull requests, so created a new one. :)

PR #213, opened 2018-05-02
Fix fftw_complex usage

ginkage:
See issue #211 for the back story.


karlstav:
looks good. Thanks again for spotting this!

Issue #112, opened 2016-10-18
Adding graphical support


nikp123:
Can you add GUI support to CAVA. I know it's supposed to be console-based, but I got a Xlib/X11 version working here: https://github.com/nikp123/cava-gui

EDIT: Sorry if I am doing anything wrong, still learning how github works :)

karlstav:
@nikp123 I would really like to check out your fork, but it segfaults :confused:  (I'm running linux mint 18) so I guess it still needs some work.

Making some more graphics has been on the to-do-list for a while. Maybe also some kind of opengl thing. But most importantly it needs to be optional, so if someone doesn't want it or can't install it the config script will disable it (same way as is done with ncurses and pulseaudio). So if you can make it as a separate module (along with the other output modules), that would be great. Also I think that it would be best to have a separate binary since it's sort of a different application. If you can make all that I can create a separate branch for it here.

OR you could keep this as a separate project, cava was after all meant to be a terminal-based application (as the C stands for console).

btw you are not doing anything wrong :wink:, but github or git can be very difficult to learn, I have used [this](https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging) and [this](http://ohshitgit.com/) article a lot.

nikp123:
@karlstav I made them as availble output modes, and also reconfigured the configuration scripts. Check your pull requests :)

nikp123:
@karlstav I guess the issue with bars is that you made over 200 bars, and that causes a segfault. Beware, geometry in the GUI mode is pixel based. Change the bar width to the desired ammount of pixels.

nikp123:
@karlstav Also the old repo is destroyed, and I made a fork as you can see.

karlstav:
hmm I get an xwindow now but no bars, I aslo get this error in the terminal:

```
XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":0"
      after 5364 requests (5364 known processed) with 0 events remaining.
```

still looks as if it needs a lot of work... I have to admit that my time is limited at the moment, so I think it's best to use your own repo for development.

nikp123:
[...]

PR #226, opened 2018-07-01
[#112] Added Xorg, SDL2 and GL support

nikp123:


nikp123:
It also contains some bugfixes.

karlstav:
I will have to do some testing before I can merge this. I also have to do some considerations on whether or not I can maintain all this extra functionality or if it diverts to much from the original project. 

Right now time is limited, I will get back to this.

nikp123:
@karlstav no problem, I was thinking if you wanted these features that I would merge them with your repository. But I thought to keep my fork seperate, because I would like to experiment with a lot of functionality (which breaks legacy configurations). If you do want to merge stuff, please list them below so I could sort it out.

nikp123:
I think that maintaining graphical modes for cava is going to be hard (as I've spent hours merging incompatible code), so I'm thinking that it's a better idea to keep this as a fork (probably should rename it to gava just to avoid conflicts codewise and configwise). Meanwhile you can still merge the non-graphics code I have, such as portaudio.

karlstav:
@nikp123 agreed, I have tried to look trough your code from time to time, but I haven't been able to grasp it completely. So it is probably best you maintain it yourself.

Also there is already an openGL implementation called [glava](https://github.com/wacossusca34/glava), so maybe gava is not the best name. What about xava?

nikp123:
@karlstav While I like the name, it implies that it's for the X windowing system. Then again the code is really meant for Linux so it's appropriate. I mean I can take it, but I'm just trying to avoid confusion.

PR #230, opened 2018-08-03
Enabled multicolored gradient

iamsubhranil:
Now, the option to configure gradient with more than 2 colors is
supported. To use this, write the following in your config :
```
gradient = 1
gradient_count = <num>
gradient_color_1 = <col>
gradient_color_2 = <col>
gradient_color_3 = <col>
.
.
.
gradient_color_<num> = <col>
```
Replace `<num>` with the number of colors (2 <= colors <= 8), and replace
 `<col>`s with your colors of choice. You must explicitly provide all
 colors for the gradient to work, otherwise appropriate errors will be
 shown.

Also, the default value for `gradient_count` is 2 when absent. Hence all old configs will be valid too.


karlstav:
works beautifully, thanks!

PR #234, opened 2018-09-07
Feature: reload colors only

dgrisham:
I have a use-case where I'm setting the visualizer's gradient colors based on album art. To do so, I was using the `SIGUSR1` signal to reload the config every time the album changed. However, the full config reload has more visual effects than simply changing the color -- the visualizer resets completely, and the bars have to re-find their optimal size, etc. To make the transition smoother, I made it so that `SIGUSR2` reloads the colors only. I also bound key `g` to this feature, which can be useful for users testing their color configs as well.

Not sure if this is something you want to support, but I thought I'd make a PR since I made the change :)

### DIFF
--- a/config.c
+++ b/config.c
@@ -1,3 +1,4 @@
+#define MAX_ERROR_LEN 1024
 
 double smoothDef[64] = {0.8, 0.8, 1, 1, 0.8, 0.8, 1, 0.8, 0.8, 1, 1, 0.8,
 					1, 1, 0.8, 0.6, 0.6, 0.7, 0.8, 0.8, 0.8, 0.8, 0.8,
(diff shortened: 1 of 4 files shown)


dgrisham:
I think it makes sense have this reload the rest of the colors as well (foreground/background) -- it makes sense in both use cases, i.e. 1. testing colors and doing quick reloads and 2. dynamically updating colors for theming/etc. purposes. I'll implement this as well, assuming it goes as smoothly as I expect.

dgrisham:
Hey @karlstav, just pushed the error handling update. Everything now works as we discussed in https://github.com/karlstav/cava/issues/235. Let me know what you think -- I tried to make it both robust and clean, which led to [the `write_errorf()` function that handles most of the error updating](https://github.com/karlstav/cava/pull/234/files#diff-e3525ca4a1a338ed0115251c17b21abeR30).

I initially had a simpler error handling setup, but ended up needing to change it a bit because of a few lines like [this one](https://github.com/karlstav/cava/pull/234/files#diff-e3525ca4a1a338ed0115251c17b21abeR81), which appends an additional error message to the one [here](https://github.com/karlstav/cava/pull/234/files#diff-e3525ca4a1a338ed0115251c17b21abeR45). Because of this, I needed a way to maintain the size of the string currently written to the error message, which led to the `error_s` struct (which has a `length` field). I figured I'd solve the problem generally, which ended up at the aforementioned `write_errorf()` which also ensures that the `error.message` buffer is not written past its end.

So, generally, the benefits of `write_errorf()` are:

1. Error message is always updated safely, and relevant state (the current length) is maintained behind the scenes.
2. Easy to update in the future if additional error output is added (programmer doesn't have to worry about buffer overruns).
3. `write_errorf()` behaves as `printf()`/etc., so it's very straightforward/familiar to use.

I haven't written C in awhile and kind of just stumbled across this approach, but I think it turned out pretty clean. Let me know what you think.

config.c
```diff
@@ -291,9 +347,8 @@ if (configPath[0] == '\0') {
 		configHome = getenv("HOME");
 		if (configHome != NULL) {
 			sprintf(configPath,"%s/%s/%s/", configHome, ".config", PACKAGE);
-		} else {
```
karlstav: I don't think you should have removed this line, now it will always fail with ERR_HOMELESS, unless you have XDG_CONFIG_HOME defined, wich some systems don't, like debian dabes systems.

dgrisham: Ah, good catch. That was accidental, I'll add it back in.

dgrisham:
@karlstav Fixed that mistake.

karlstav:
@dgrisham looks good, do you want to add aline or two in the readme as well:

https://github.com/karlstav/cava#configuration

dgrisham:
@karlstav updated!

Issue #243, opened 2018-12-03
support for Squeezelite


dnalor:
I've written an input module for [Squeezelite](https://en.wikipedia.org/wiki/Squeezelite).
As Squeezelite can export it's audio data as shared memory, it's a simple solution without external dependencies.
If this is of interest, I can contribute it to the project.

karlstav:
cool, just make a PR. I guess I should make some kind of input module contribution guideline. But just make it consistent with the other modules.

PR #244, opened 2018-12-06
Input shmem

dnalor:
as discussed in #243


cava.c
```diff
@@ -441,6 +448,13 @@ as of 0.4.0 all options are specified in config file, see in '/home/username/.co
 	}
 	#endif
 
+	#ifdef SHMEM
+	if (p.im == 5) {
+		thr_id = pthread_create(&p_thread, NULL, input_shmem, (void*)&audio);
+		//audio.rate = 44100;
```
dnalor: not quite sure about this

karlstav:
looks good, I will test it later. Can you also add something in the example config file?

karlstav:
I don't have the chance to test it with squeezelite just now. But build and run looks good. After some more testing it can probably be enabled by default.

Issue #275, opened 2019-05-27
About MPD FIFO rate


starmood:
Is that possible to use different FIFO format when using CAVA with MPD. Right now, if I try something else than "44100", the displaying is incorrect. 
But unless the original music file is 44100, the FIFO output of MPD will be resampled . And it will cause a lot CPU resouce on low performance platform, like RasPi. 
Any solution that I can use CAVA with non-resampled FIFO, or I can somehow reduce the cost of re-sampling?

Thanks.

karlstav:
The sample rate of the fifo input is hard coded. But maybe there could have been a config option for it. I don't know if it's possible to detect it from the fifo stream automatically. Maybe it's possible to request it from MPD via some API. That would have been cool. Then we could also have like track name displayed when ever there is a new song.

starmood:
I am not expert on this. So you mean if you can get the FIFO rate, it's possible to decode it, it's not neccessary with 44.1k?

karlstav:
I thought I responded to this right away, but must have forgotten. 

Yes, if cava can get the fifo rate from mpd then cava can use that instead of 44.1k. That is how the alsa input works and I should attempt a similar solution for mpd.

karlstav:
doing this manually is probably a good enough solution as you would never really now the rate of a fifo.

PR #291, opened 2019-12-17
Allow FIFO input sample rate to be configurable

quantum5:
Refs #275.

While it was mentioned that the ability to acquire this information from MPD would be useful, allowing it to be configured explicitly could help when FIFO is used for non-MPD sources.

For example, I am currently running Cava inside Windows Subsystem for Linux with a program on the Windows side feeding the capture of the system audio output into Cava, and it would be really useful to be able to handle 48 kHz audio as that is the default format on Windows.


karlstav:
nice, this could actually be a generic sample rate setting (not just for fifo) since pulseaudio and portaudio also just use a hardcoded rate.

PR #299, opened 2019-12-20
Use NDEBUG for debugging and add debug print helper

Xyene:
This PR is split up in two parts.

First, usage of `DEBUG` is remapped to `NDEBUG`, such that the behaviour of assert(3) is as expected with --enable-debug=false (i.e. asserts get disabled, where they previously did not).

Second, instances of `printf` guarded by `#ifdef DEBUG` blocks have been refactored into a dedicated `debug` macro to make the code a bit easier to follow. `debug` is a no-op when `NDEBUG` is set.

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -31,6 +31,8 @@
 #include <dirent.h>
 #include <ctype.h>
 
+#include "debug.h"
+
 #ifdef NCURSES
 #include "output/terminal_ncurses.h"
 #include "output/terminal_ncurses.c"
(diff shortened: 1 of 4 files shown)


karlstav:
sounds good. The debug option was added in the very beginning and wasn't exactly done by the book.

karlstav:
hmm I miss the table output of all defined here:

https://github.com/karlstav/cava/blob/db7f8d1ccbe98abd022934aa209bbf93dd3be3d2/cava.c#L991

why are some things still DEBUG while some are NDEBUG?

PR #292, opened 2019-12-18
Use clang-format to ensure a consistent style

quantum5:
A proposal to ensure a consistent style in the codebase. Please feel free to suggest changes to the style in `.clang-format` or reject the PR.

PR #300, opened 2019-12-20
Rework output/terminal_ncurses.c to be more performant

Xyene:
By scanning through the updates row-by-row rather than column-by-column, a lot of color-changing updates are saved when using a gradient. This seems to significantly reduce tearing when running Cava through urxvt on an X11 server.

(Apologies for the otherwise-large diff; my editor normalized the file to be tabbed-only. If this is a problem I can undo that.)

### DIFF
--- a/output/terminal_ncurses.c
+++ b/output/terminal_ncurses.c
@@ -71,79 +71,79 @@ char* const bg_color_string, int predef_fg_color, int predef_bg_color, int gradi
 	start_color();
 	use_default_colors();
 
-    getmaxyx(stdscr, *height, *width);
+	getmaxyx(stdscr, *height, *width);
 	clear();
 
-    NCURSES_COLOR_T color_pair_number = 1;
+	NCURSES_COLOR_T color_pair_number = 1;
 
-    NCURSES_COLOR_T bg_color_number;
-    bg_color_number = change_color_definition(0, bg_color_string, predef_bg_color);
+	NCURSES_COLOR_T bg_color_number;
+	bg_color_number = change_color_definition(0, bg_color_string, predef_bg_color);
 
-    if (!gradient) {
+	if (!gradient) {
 
-	    NCURSES_COLOR_T fg_color_number;
-        fg_color_number = change_color_definition(1, fg_color_string, predef_fg_color);
+		NCURSES_COLOR_T fg_color_number;
+		fg_color_number = change_color_definition(1, fg_color_string, predef_fg_color);
 
-        init_pair(color_pair_number, fg_color_number, bg_color_number);
+		init_pair(color_pair_number, fg_color_number, bg_color_number);
 
-    } else if (gradient) {
+	} else if (gradient) {
 
-        // 0 -> col1, 1-> col1<=>col2, 2 -> col2 and so on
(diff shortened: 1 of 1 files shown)


karlstav:
that's ok I'm planning to do some kind of clang formatting  ref #292 

This did however screw up the bars in tty, not sure what happened. Can you take a look @Xyene ?

Xyene:
I can take a look. I only have a pty to test in, so hoped that the behaviour would be the same... guess not. What does it look like on a tty?

Xyene:
Oh, I think I see it. If you change '0' to '1' here:

https://github.com/karlstav/cava/pull/300/files#diff-b5cc6870fd48981b2eb151abf58d51e8R224

Does it get fixed?

karlstav:
introduced here (I think) when f[n] is `1` as with no sound, it no longer uses the 1/8 bar... more like 2/8 or something. So the bars are a bit thicker when there is no sound, can you take a look @Xyene ?

karlstav:
there should proabably be a minus one here:

https://github.com/karlstav/cava/blob/456bfc0919d753ce2bb4e006b429f9348d9a6dfb/output/terminal_ncurses.c#L221

Xyene:
You're right, that should be a - 1... my font was small enough that I didn't notice the difference 😅 

It doesn't seem to segfault for me, though. I'll try it out on another machine.

karlstav:
it segfaults ones a bar gets up to one whole cell. I suspect that this will always be true:

https://github.com/karlstav/cava/blob/456bfc0919d753ce2bb4e006b429f9348d9a6dfb/output/terminal_ncurses.c#L219

Xyene:
I think the `else` of that `if` does get executed, since removing it results in this:

image

I believe it should execute any time a bar grows by more than one cell in a single frame.

I've created https://github.com/karlstav/cava/pull/323 just so we're on the same page and so it's easier to see the diff.

karlstav:
I tested the changes in that PR and it does not look right at all. A lot of flickering.

karlstav:
adding this `if (f[bar] % 8 == 0) continue;` on line 221 in addition to #323 solves the issue. I think that the problem is that you can be on "top" of the bar without actually having any more "parts of eight" bars to to draw.

Xyene:
Ahhhh, I think I see... the fact that f[bar] starts off at 1 rather than 0 threw off these lines too:

```diff
             int cur_col = bar * bar_width + bar * bar_spacing + rest;
-            int f_cell = f[bar] / num_bar_heights;
-            int f_last_cell = flastd[bar] / num_bar_heights;
+            int f_cell = (f[bar] - 1) / num_bar_heights;
+            int f_last_cell = (flastd[bar] - 1) / num_bar_heights;
 
             if (f_cell >= y) {
                 int bar_step;
```

The cell should be 0 for the values [1, 8], but before the cell would be 1 when the value is 8, which is... incorrect. So it was trying to redraw cells that it didn't have to, exactly when `f[bar] % 8 == 0`. Can you give that a try and see if it works?

karlstav:
yea that look good. The thing was that I wanted a thin line at the bottom when there was silence instead  of just nothing. That might been the cause of the confusion.

PR #302, opened 2019-12-20
Missed two DEBUG to NDEBUG conversions

Xyene:
Looks like I missed one in https://github.com/karlstav/cava/pull/299, even though I had done a grep for DEBUGs. Sorry about that.

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -923,7 +923,7 @@ as of 0.4.0 all options are specified in config file, see in '/home/username/.co
 
 			}
 			 else { //**if in sleep mode wait and continue**//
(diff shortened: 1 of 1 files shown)


karlstav:
I am still not getting it to print. Does this : `#ifndef NDEBUG` really work?

Xyene:
That's odd, it should... I'm getting this output as of this PR:

image

Which seems to match the output from 04eabc7d8da04b96e9987f7f27047d69067dfd48, prior to the first debug change:

image

karlstav:
ahh ok, something with my terminal probably.

PR #304, opened 2019-12-20
ncurses: tentatively fix tty output

Xyene:
This PR tentatively fixes the issue reported in #300. I don't have a real TTY to test on, so I'm not sure if this works, though.


karlstav:
that did the trick, thanks for all you contributions @Xyene !

Issue #306, opened 2019-12-20
Support compiling using AddressSanitizer


quantum5:
Add a `--enable-asan` flag to enable compilation with AddressSanitizer. This will help track down any buffer overflows, memory leaks, and other assorted memory-related bugs in the program.

quantum5:
I think ASan is ready now.

PR #309, opened 2019-12-20
Reduce ASan leak reports by moving buffers to global scope

quantum5:
Refs #306.

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -100,6 +100,17 @@ int should_reload = 0;
 // whether we should only reload colors or not
 int reload_colors = 0;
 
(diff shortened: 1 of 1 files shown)


quantum5:
It appears that most of these buffers are already freed at the end of main? I think I'll need some more investigation.

quantum5:
It appears that Cava simply exits without freeing the buffers when `q` is pressed:

```c
case 'q':
	cleanup();
	return EXIT_SUCCESS;
```

In this case I think this PR is fine to merge as is.

PR #310, opened 2019-12-20
Fix memory leak in load_config with strdup and iniparser_freedict

quantum5:
Memory leaks every time settings are reloaded.

Refs #306.


karlstav:
somehow this commit messed up the colors in my terminal, like ncurses is no longer able to change color definitions. can you please investigate @quantum5 ?

karlstav:
it is the ` iniparser_freedict(ini);` I don't remember why, but somehow it prevents ncurses from changing the color definitions.

PR #311, opened 2019-12-20
Use loop exit condition in place of always true

Xyene:
This PR proposes a minor change to when loops are exited, in hopes that it makes tracking when resources are freed easier to track.


karlstav:
agreed.

PR #312, opened 2019-12-20
Use free to free out_*_l/r instead of fftw_free

quantum5:
These are allocated with calloc and so should be freed with free.


quantum5:
Hmm... a further examination suggests that the allocation should be done with `fftw_alloc_complex` instead to allow fftw to use SIMD. See http://www.fftw.org/fftw3_doc/Memory-Allocation.html

karlstav:
you're right, do you want to make the changes?

quantum5:
Yes, I will go and make the changes.

PR #318, opened 2019-12-21
Compile Cava as multiple object files

### LINKED ISSUE #317: Compile with multiple object files
Currently, Cava has multiple source files, but they are all `#include`d into `cava.c`, which is not a common or recommended practice. We should compile each source file to a separate object file.

quantum5:
Fixes #317.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -10,7 +10,8 @@ M_CPPFLAGS = -DSYSTEM_LIBINIPARSER=@SYSTEM_LIBINIPARSER@
 
 bin_PROGRAMS = cava
 cava_SOURCES = cava.c input/common.c input/alsa.c input/fifo.c input/portaudio.c input/pulse.c \
-               input/sndio.c input/shmem.c
+               input/sndio.c input/shmem.c output/terminal_bcircle.c output/terminal_ncurses.c \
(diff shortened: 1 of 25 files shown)


karlstav:
feel free to rebase/resolve and thanks again for all the contributions @quantum5 and @Xyene! This is stuff that I would have never had time to do on my own.

quantum5:
@karlstav this is ready now.

Xyene:
@karlstav no problem, happy to help 😄 

For some context: both myself and @quantum5 [have a weird use-case for cava](https://quantum2.xyz/2019/12/16/desktop-audio-visualizer-i3-cava-wsl/) (hence the 24-bit FIFO input, etc.)... on one hand, it looks kinda cool:

image

...on the other, it's only practical (or... not impractical?) if cava + urxvt use as little CPU as possible 😉. In our case at least, the ncurses change + making FFTW aligned for SIMD dropped usage from 20~30% to 0-4%.

karlstav:
look nice, for me CPU usage for cava with a tiny tiny urxvt window is around 5% while full screen terminal is around 10%. I think a full screen window had much higher cpu usage before your much appreciated contributions!

PR #320, opened 2019-12-21
Refactor input method magic numbers into enum

Xyene:
(This PR is based off of #318, I will rebase it if/when that one is approved.)

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -268,7 +268,6 @@ as of 0.4.0 all options are specified in config file, see in '/home/username/.co
 
     char ch = '\0';
     int bars = 25;
-    char supportedInput[255] = "'fifo'";
     int sourceIsAuto = 1;
     double smh;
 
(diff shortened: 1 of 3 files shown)


karlstav:
did I pull the wrong request now?

Xyene:
Nope, everything's good, I've rebased this change and https://github.com/karlstav/cava/pull/321.

karlstav:
hmm I get no output after this, bars act like if silence.

Xyene:
@karlstav what input/output methods do you have configured?

karlstav:
none, seems like pulseaudio is no longer defualt.

karlstav:
is `fifo`default now? I think it would be better if pulseaudio is default. If you then have pulseaudio installed in your system it would be a simple install and run to capture audio straight from your system (since the pulseaudio input module is able to do that). Also to have portaudio to default before fifo would be convenient for macos users

Xyene:
Agreed, `fifo` should be picked last. I thought I had maintained the correct order, but looks like I mucked something up. I'll submit a follow-up to fix it.

Issue #293, opened 2019-12-18
Support 24-bit audio as configuration option


Xyene:
Windows has support for 24-bit audio streams, which could hypothetically be used with Cava's FIFO input running in WSL. #291 improved the Windows situation a bit, however, Cava appears to only support 16-bit audio.

(I am happy to implement this myself, but am opening this issue as per [CONTRIBUTING.md](https://github.com/karlstav/cava/blob/master/CONTRIBUTING.md).)

karlstav:
actually the code for using any sample format exists in the alsa input module:

https://github.com/karlstav/cava/blob/9610991b3a90ba676a759f58967164a52b5e33d8/input/alsa.c#L132

This was actually the only thing that was used in the beginning. I haven't tested it in a while, but I think it still works. The comment about only one channel and biggest octet does not apply anymore.

I think the most practical approach would be to move the whole case switch for the sample format here:

https://github.com/karlstav/cava/blob/9610991b3a90ba676a759f58967164a52b5e33d8/input/fifo.c#L30

this function is used by all the audio input modules to pass audio to the input buffer used by FFTW.

Feel free to implement and create a PR.

Xyene:
This has been implemented.

PR #321, opened 2019-12-21
Add support for 24- and 32-bit FIFO input

Xyene:
Refs #293.

This also fixes a possible bug with 16-bit audio: if `read` spuriously returns having read less than a multiple of 2 bytes, the stream could get unsynced forever. I think this is unlikely for 16-bit samples, but happens *very* frequently with 24-bit samples. (This is fixed by simply waiting until the entire sample buffer is full before proceeding.)

(This PR is based off of #318, I will rebase it if/when that one is approved.)

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -414,6 +414,7 @@ as of 0.4.0 all options are specified in config file, see in '/home/username/.co
             // starting fifomusic listener
             thr_id = pthread_create(&p_thread, NULL, input_fifo, (void *)&audio);
             audio.rate = p.fifoSample;
+            audio.format = p.fifoSampleBits;
             break;
 #ifdef PULSE
         case INPUT_PULSE:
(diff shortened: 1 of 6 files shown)


karlstav:
nice, could you also document the new config param in the example config file?

karlstav:
got this warning though:
```
input/common.c: In function ‘reset_output_buffers’:
input/common.c:4:5: warning: implicit declaration of function ‘memset’ [-Wimplicit-function-declaratio ]
    4 |     memset(data->audio_out_bass_r, 0, sizeof(int16_t) * data->FFTbassbufferSize);
      |     ^~~~~~
input/common.c:4:5: warning: incompatible implicit declaration of built-in function ‘memset’
input/common.c:2:1: note: include ‘<string.h>’ or provide a declaration of ‘memset’
    1 | #include "input/common.h"
  +++ |+#include <string.h>
    2 | 
```

does string.h need to be included in this file as well? it is included in cava,c.

Xyene:
I noticed and fixed this in https://github.com/karlstav/cava/pull/326, so it should be OK now.

PR #330, opened 2019-12-21
Update example config with FIFO sample rate and bit depth

Xyene:
As requested in #321. Also removed some trailing spaces in the config file.


karlstav:
nice. as mention earlier the config options for sample rate and format could also apply for all other input modules except alsa, as I believe they are hardcoded in all those as well.

PR #338, opened 2020-03-01
add configure options to disable optional features

antiprism:
When building a cava binary on one machine to be used on another the features included, and hence the dependencies for the binary, are determined by the libraries installed on the development machine. Rather than create a build environment that only has access to the dependencies that are wanted for the final binary, this patch allows optional features to be disabled by configure options.

Strange issue...

When building for a Pi Zero on Raspbian with clang  (version 7.0.1-8+rpi3) , either building on the Pi Zero itslf or on a Pi 4 with `-march=armv6 -mfpu=vfp` and everything disabled except for input-alsa , the binary did not work correctly for fifo input and raw output, with levels rising quickly to maximum
```
pi@moode:~/cava $ cava -p conf3.txt 
open file /dev/stdout for writing raw ouput
9;10;255;79;72;41;32;16;16;17;
29;40;894;320;266;135;111;57;62;68;
61;109;1000;870;709;350;291;151;167;185;
155;285;1000;1000;1000;841;687;370;392;437;
369;700;1000;1000;1000;1000;1000;870;915;1000;
858;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
1000;1000;1000;1000;1000;1000;1000;1000;1000;1000;
```
It worked fine when built with gcc

It seems to be related to some kind of interaction between clang, the alsa output feature, and the libraries "-ldl -ltinfo".

The issue occurs when running on the Pi Zero (binary built on the Pi Zero or on a Pi 4, clang version 7.0.1-8+rpi3). I didn't test on the Pi 4 itself. I tested on an Intel PC and the native clang (version 6.0.0-1ubuntu2) and didn't encounter the issue.

Here are some example configures that trigger the issue or not.

Triggers the issue (everything disabled, LDFLAGS used): 
```CC=clang CPPFLAGS="-march=armv6 -mfpu=vfp -W -Wall" LDFLAGS="-ldl -ltinfo" ./configure  --disable-input-alsa --disable-input-pulse --disable-input-portaudio --disable-input-sndio --disable-output-ncurses```

Doesn't trigger the issue (everything disabled, LDFLAGS not used)
```CC=clang CPPFLAGS="-march=armv6 -mfpu=vfp -W -Wall"  ./configure  --disable-input-alsa --disable-input-pulse --disable-input-portaudio --disable-input-sndio --disable-output-ncurses```

Doesn't trigger the issue (output-alsa enabled, LDFLAGS used): 
```CC=clang CPPFLAGS="-march=armv6 -mfpu=vfp -W -Wall" LDFLAGS="-ldl -ltinfo" ./configure --disable-input-pulse --disable-input-portaudio --disable-input-sndio --disable-output-ncurses```

Triggers the issue (output-alsa enabled, LDFLAGS not used):
```CC=clang CPPFLAGS="-march=armv6 -mfpu=vfp -W -Wall"  ./configure --disable-input-pulse --disable-input-portaudio --disable-input-sndio --disable-output-ncurses```

Doesn't trigger the issue (output-alsa enabled, LDFLAGS not used, built with gcc):
```CC=gcc CPPFLAGS="-march=armv6 -mfpu=vfp -W -Wall"  ./configure --disable-input-pulse --disable-input-portaudio --disable-input-sndio --disable-output-ncurses```

I have just cut and pasted the above commands and repeated the builds with the same results. I don't have an explanation for the behaviour.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -77,63 +77,94 @@ AC_CHECK_HEADER([alloca.h], [CPPFLAGS="$CPPFLAGS -DHAVE_ALLOCA_H"])
 dnl ######################
 dnl checking for alsa dev
 dnl ######################
-AC_CHECK_LIB(asound, snd_pcm_open, have_alsa=yes, have_alsa=no)
-    if [[ $have_alsa = "yes" ]] ; then
-      LIBS="$LIBS -lasound"
-      CPPFLAGS="$CPPFLAGS -DALSA"
-    fi
+AC_ARG_ENABLE([input_alsa],
+  AS_HELP_STRING([--disable-input-alsa],
+    [do not include support for input from alsa streams])
+)
 
-    if [[ $have_alsa = "no" ]] ; then
-      AC_MSG_NOTICE([WARNING: No alsa dev files found building without alsa support])
-    fi
+AS_IF([test "x$enable_input_alsa" != "xno"], [
+  AC_CHECK_LIB(asound, snd_pcm_open, have_alsa=yes, have_alsa=no)
+  if [[ $have_alsa = "yes" ]] ; then
+    LIBS="$LIBS -lasound"
+    CPPFLAGS="$CPPFLAGS -DALSA"
+  fi
+  if [[ $have_alsa = "no" ]] ; then
+    AC_MSG_NOTICE([WARNING: No alsa dev files found building without alsa support])
+  fi],
+  [have_alsa=no]
+)
 
-    AM_CONDITIONAL([ALSA], [test x$have_alsa = xyes])
+AM_CONDITIONAL([ALSA], [test "x$have_alsa" = "xyes"])
 
 
 dnl ######################
 dnl checking for pulse dev
 dnl ######################
-AC_CHECK_LIB(pulse-simple, pa_simple_new, have_pulse=yes, have_pulse=no)
-    if [[ $have_pulse = "yes" ]] ; then
-      LIBS="$LIBS -lpulse-simple -lpulse"
-      CPPFLAGS="$CPPFLAGS -DPULSE"
-    fi
+AC_ARG_ENABLE([input_pulse],
+  AS_HELP_STRING([--disable-input-pulse],
+    [do not include support for input from pulseaudio])
+)
 
-    if [[ $have_pulse = "no" ]] ; then
-      AC_MSG_NOTICE([WARNING: No pusleaudio dev files found building without pulseaudio support])
-    fi
+AS_IF([test "x$enable_input_pulse" != "xno"], [
+  AC_CHECK_LIB(pulse-simple, pa_simple_new, have_pulse=yes, have_pulse=no)
+  if [[ $have_pulse = "yes" ]] ; then
+    LIBS="$LIBS -lpulse-simple -lpulse"
+    CPPFLAGS="$CPPFLAGS -DPULSE"
+  fi
 
-    AM_CONDITIONAL([PULSE], [test x$have_pulse = xyes])
+  if [[ $have_pulse = "no" ]] ; then
+    AC_MSG_NOTICE([WARNING: No pusleaudio dev files found building without pulseaudio support])
+  fi],
+  [have_pulse=no]
+)
+
+AM_CONDITIONAL([PULSE], [test "x$have_pulse" = "xyes"])
 
 dnl ######################
 dnl checking for portaudio dev
 dnl ######################
-AC_CHECK_LIB(portaudio, Pa_Initialize, have_portaudio=yes, have_portaudio=no)
-    if [[ $have_portaudio = "yes" ]] ; then
-      LIBS="$LIBS -lportaudio"
-      CPPFLAGS="$CPPFLAGS -DPORTAUDIO"
-    fi
+AC_ARG_ENABLE([input_portaudio],
+  AS_HELP_STRING([--disable-input-portaudio],
+    [do not include support for input from portaudio])
+)
 
-    if [[ $have_portaudio = "no" ]] ; then
-      AC_MSG_NOTICE([WARNING: No portaudio dev files found building without portaudio support])
-    fi
+AS_IF([test "x$enable_input_portaudio" != "xno"], [
+  AC_CHECK_LIB(portaudio, Pa_Initialize, have_portaudio=yes, have_portaudio=no)
+  if [[ $have_portaudio = "yes" ]] ; then
+    LIBS="$LIBS -lportaudio"
+    CPPFLAGS="$CPPFLAGS -DPORTAUDIO"
+  fi
 
-    AM_CONDITIONAL([PORTAUDIO], [test x$have_portaudio = xyes])
+  if [[ $have_portaudio = "no" ]] ; then
+    AC_MSG_NOTICE([WARNING: No portaudio dev files found building without portaudio support])
+  fi],
+  [have_portaudio=no]
+)
+
+AM_CONDITIONAL([PORTAUDIO], [test "x$have_portaudio" = "xyes"])
 
 dnl ######################
 dnl checking for sndio dev
 dnl ######################
-AC_CHECK_LIB(sndio, sio_open, have_sndio=yes, have_sndio=no)
-    if [[ $have_sndio = "yes" ]] ; then
-      LIBS="$LIBS -lsndio"
-      CPPFLAGS="$CPPFLAGS -DSNDIO"
-    fi
+AC_ARG_ENABLE([input_sndio],
+  AS_HELP_STRING([--disable-input-sndio],
+    [do not include support for input from sndio])
+)
 
-    if [[ $have_sndio = "no" ]] ; then
-      AC_MSG_NOTICE([WARNING: No sndio dev files found building without sndio support])
-    fi
+AS_IF([test "x$enable_input_sndio" != "xno"], [
+  AC_CHECK_LIB(sndio, sio_open, have_sndio=yes, have_sndio=no)
+  if [[ $have_sndio = "yes" ]] ; then
+    LIBS="$LIBS -lsndio"
+    CPPFLAGS="$CPPFLAGS -DSNDIO"
+  fi
 
-    AM_CONDITIONAL([SNDIO], [test x$have_sndio = xyes])
+  if [[ $have_sndio = "no" ]] ; then
+    AC_MSG_NOTICE([WARNING: No sndio dev files found building without sndio support])
+  fi],
+  [have_portaudio=no]
+)
+
+AM_CONDITIONAL([SNDIO], [test "x$have_sndio" = "xyes"])
 
 dnl ######################
 dnl checking for math lib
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by karlstav** [comment]
looks good, would you add a note on how to use this in the readme? 

Regarding your issue, can it be using the alsa input when you are encountering the strange input?

**Comment by antiprism** [comment]
Thanks for including this.

Here is a note, at the end of the following paragraph, which you could add into the README

Only FFTW and the other build tools are actually required for CAVA to compile, but this will only give you the ability to read from fifo files. To more easly grab audio from your system pulseaudio, alsa, sndio or portaudio dev files are recommended (depending on what audio system you are using). Not sure how to get the pulseaudio dev files for other distros than debian/ubuntu or if they are bundled in pulseaudio. If you have a recommended component installed, but do not wish to use it (perhaps if building a binary on one machine to be used on another), then the corresponding feature can be disabled during configuration (see `configure --help` for details).

I plan to have a look at what is happenning with the issue. Maybe it is just a Pi Zero thing. I came across some stack corruption in another project the other day that seemed to be produced by apparently correct alsa *_alloca() calls. Changing these to *_malloc() calls fixed the issue.

**Comment by antiprism** [comment]
It seems that the issue can be removed by building with CFLAGS="-O0"

It probably has the same root as the issue I had with the other project, where I could work around the stack corruption by building with -O0:
Issue: https://github.com/BlokasLabs/amidiauto/issues/1
My fix: https://github.com/antiprism/amidiauto/commit/d749bac48d1a989245047fd25d0c60e8b9dbff3e

For cava, changing the *_alloca in input/alsa.c won't fix the issue, as this file isn't compiled when using --disable-input-alsa but the issue can still be triggered.

I don't think there is much value in pursuing this unless it affects more people.

**Comment by karlstav** [comment]
I thought `-O0` was default? At least that is what [this](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html) says. I remember testing out `-O3` for performance, but I don't think i committed it. 

I see now in the build log that it actually uses `-O2` (Optimize even more), now where does that sneak in!?

**Comment by antiprism** [comment]
The CFLAGS default comes from autoconf's AC_PROG_CC

   https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/C-Compiler.html

**Comment by karlstav** [comment]
right, it looks like O2 is sort of the recommended value (compile time is not an issue anyways).

You could check out the latest commit. I removed some unnecessary buffers used between audio source and FFTW (I had planed to do this anyways)

**Comment by antiprism** [comment]
I tried the latest changes, but they don't affect the build issue.

**Comment by karlstav** [comment]
ok it was a long shot. I support the theory of stack corruption. However I can't rule out the possibility that this issue might also be related to some weakness in the code. 

Out of curiosity what is you configuration? Is autosens enabled? And what is you expected output when the issue occurs?

**Comment by antiprism** [comment]
The issue could be some kind of undefined behaviour occuring in the cava code, but might be difficult to track down.

Here is a warning, but this code isn't built when everything is disabled, and so isn't the issue.
```
input/portaudio.c:29:44: warning: implicit conversion from 'int' to 'int16_t'
      (aka 'short') changes value from 32768 to -32768 [-Wconstant-conversion]
    int16_t silence_buffer[PORTBUFSIZE] = {SAMPLE_SILENCE};
                                          ~^~~~~~~~~~~~~~
input/portaudio.c:7:24: note: expanded from macro 'SAMPLE_SILENCE'
#define SAMPLE_SILENCE 32768
                       ^~~~~
```
This is my configuration
```
[general]
framerate = 15
bars = 21

[input]
method = fifo
source = /tmp/mpd_oled_fifo

[output]
method = raw
data_format = binary
channels = mono
raw_target = /tmp/cava_fifo_263
bit_format = 8bit
```
The input is music and the output should be a typical music spectum.

**Comment by karlstav** [comment]
maybe be some uninitialized variables or buffers not properly allocated or stack corruption or a combination? As you say, hard to track down. As this only happens when building with clang instead of gcc it is a pretty rare case and maybe not worth spending much time on.

**Comment by antiprism** [comment]
I don't think it is worth spending time on unless you receive more reports of the issue.

I did spot in input/fifo.c there is
```
uint8_t buf[SAMPLES_PER_BUFFER * bytes_per_sample];
uint16_t *samples =
          bytes_per_sample == 2 ? (uint16_t *)&buf : calloc(SAMPLES_PER_BUFFER,   sizeof(uint16_t));
...
if (bytes_per_sample == 2) {
          free(samples);
}
```
Which looks like it should be ```if (bytes_per_sample != 2)``` (not that this helps with issue).

**Comment by karlstav** [comment]
I think you are right about that, however that's only a problem on exit / config reload. 

Can the way `samples` and `buf` share address space be the source of the stack corruption?

**Comment by antiprism** [comment]
The code looks suspicious, also with the casts, but I wasn't able to identify any other problems (but they may exist!). I previously tried using the dynamically allocated memory  for both cases, but it didn't affect the issue.

**Comment by karlstav** [comment]
@quantum5 or @Xyene do you mind taking a look through this?

**Comment by quantum5** [comment]
A suggestion would to be run `./configure` with `--enable-asan`, which should help in tracking down any memory corruption, and/or `--enable-ubsan`, which should help in tracking down any undefined behaviour.

**Comment by antiprism** [comment]
Thanks for the suggestion @quantum5.

I built with
```
CC=clang CPPFLAGS="-march=armv6 -mfpu=vfp"  ./configure  --disable-input-alsa --disable-input-pulse --disable-input-portaudio --disable-input-sndio --disable-output-ncurses --enable-asan --enable-ubsan
```
and I get the following error report when running ```cava```
```
=================================================================
==15296==ERROR: AddressSanitizer: global-buffer-overflow on address 0x00172668 at pc 0x001230f4 bp 0xbe94b5b8 sp 0xbe94b5b4
READ of size 8 at 0x00172668 thread T0
    #0 0x1230f0 in main /home/pi/proj/cava/cava.c:654:25
    #1 0xb6b91714 in __libc_start_main (/lib/arm-linux-gnueabihf/libc.so.6+0x17714)

0x00172668 is located 56 bytes to the left of global variable 'default_methods' defined in 'config.c:18:19' (0x1726a0) of size 16
0x00172668 is located 0 bytes to the right of global variable 'smoothDef' defined in 'config.c:16:8' (0x172640) of size 40
SUMMARY: AddressSanitizer: global-buffer-overflow /home/pi/proj/cava/cava.c:654:25 in main
Shadow bytes around the buggy address:
  0x2002e470: f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9
  0x2002e480: f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9
  0x2002e490: f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9
  0x2002e4a0: f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9
  0x2002e4b0: f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9
=>0x2002e4c0: f9 f9 f9 f9 f9 f9 00 00 00 00 00 00 00[f9]f9 f9
  0x2002e4d0: f9 f9 f9 f9 00 00 f9 f9 f9 f9 f9 f9 00 00 00 f9
  0x2002e4e0: f9 f9 f9 f9 00 00 00 00 00 00 00 00 00 00 00 00
  0x2002e4f0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x2002e500: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x2002e510: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==15296==ABORTING
```
I added in a debugging line at cava.c line 655
```
fprintf(stderr, "(int)floor(((double)n) * smh)=%d\n", (int)flo  or(((double)n) * smh));
k[n] *= p.smooth[(int)floor(((double)n) * smh)];
```
And got
```
(int)floor(((double)n) * smh)=0
(int)floor(((double)n) * smh)=0
(int)floor(((double)n) * smh)=0
(int)floor(((double)n) * smh)=1
(int)floor(((double)n) * smh)=1
(int)floor(((double)n) * smh)=1
(int)floor(((double)n) * smh)=2
(int)floor(((double)n) * smh)=2
(int)floor(((double)n) * smh)=3
(int)floor(((double)n) * smh)=3
(int)floor(((double)n) * smh)=3
(int)floor(((double)n) * smh)=4
(int)floor(((double)n) * smh)=4
(int)floor(((double)n) * smh)=5
=================================================================
==15448==ERROR: AddressSanitizer: global-buffer-overflow on address 0x001726a8 at pc 0x00123234 bp 0xbedf7578 sp 0xbedf7574
READ of size 8 at 0x001726a8 thread T0
...
```
And so (without going through the code) it looks like the program is writing past the end of the p.smooth array
```
double smoothDef[5] = {1, 1, 1, 1, 1};
...
p->smooth = smoothDef;
```
Adrian.

**Comment by Xyene** [comment]
@antiprism @karlstav regarding the `free` -- good catch, the condition should indeed be `bytes_per_sample != 2`. In fact, I distinctly remember making that change myself, but I must've forgotten to actually send it in.

Reading that code again, I do see a potential issue, though.

```c
uint8_t buf[SAMPLES_PER_BUFFER * bytes_per_sample];
uint16_t *samples =
    bytes_per_sample == 2 ? (uint16_t *)&buf : calloc(SAMPLES_PER_BUFFER, sizeof(uint16_t));
```

The trick here was to save a `memcpy` in the common case of `bytes_per_sample` being equal to 2. However, `(uint16_t *)&buf` really relies on `buf` being 2-byte aligned, and there's nothing guaranteeing that the compiler will actually place it correctly. On x86 this'll work just fine since x86 supports unaligned memory access (albeit at a performance penalty), but I know some ARM chips definitely do not.

A quick Google suggests that unaligned accesses are fine in ARMv7, but that the Pi Zero is an ARMv6 chip. Really, the code should be written as:

```c
__attribute__((aligned(sizeof(uint16_t)))) uint8_t buf[SAMPLES_PER_BUFFER * bytes_per_sample];
```

Re: `alloca` issues, I'm curious about the output of `ulimit -a` on your system. It might be that some too-large buffers are being allocated on stack.

The ASan report looks very familiar. I remember @quantum5 running into it before, and patching it in https://github.com/karlstav/cava/commit/ccad132319fe40ac5143492dac599fd7b7a27d7a. That might've only been for the default config, though.

karlstav:
@antiprism there was indeed a bug in the default eq causing it to write after end of array, should be fixed now. will look more on the other things here later.

antiprism:
@karlstav I pulled the latest cava, and the issue with rising bar heights with certain configure options is now fixed. Thanks!

[ @Xyene for my issue with the other project, I previously tried setting the stack size limit to unlimited, but this did not help. Maybe the real issue is not the stack allocation. For reference, my limits on Moode (based Raspbian Buster, but may not be the same ulimits as stock Raspbian Buster I was using, but I recall the stack size limit is the same) on a Pi Zero are
```
pi@moode:~ $ ulimit -a
core file size          (blocks, -c) 0
data seg size           (kbytes, -d) unlimited
scheduling priority             (-e) 0
file size               (blocks, -f) unlimited
pending signals                 (-i) 3391
max locked memory       (kbytes, -l) 65536
max memory size         (kbytes, -m) unlimited
open files                      (-n) 1024
pipe size            (512 bytes, -p) 8
POSIX message queues     (bytes, -q) 819200
real-time priority              (-r) 0
stack size              (kbytes, -s) 8192
cpu time               (seconds, -t) unlimited
max user processes              (-u) 3391
virtual memory          (kbytes, -v) unlimited
file locks                      (-x) unlimited
```
]

PR #355, opened 2020-05-27
Use format string to fix build

hsheth2:
When I build cava with strict compiler settings, I get this error:
```
output/terminal_noncurses.c: In function 'draw_terminal_noncurses':
output/terminal_noncurses.c:178:81: error: format not a string literal and no format arguments [-Werror=format-security]
                         cx += snprintf(ttyline_buffer + cx, ttybuf_length - cx, ttyspacestring);
                                                                                 ^~~~~~~~~~~~~~
output/terminal_noncurses.c:180:81: error: format not a string literal and no format arguments [-Werror=format-security]
                         cx += snprintf(ttyline_buffer + cx, ttybuf_length - cx, ttybarstring[0]);
                                                                                 ^~~~~~~~~~~~
output/terminal_noncurses.c:183:40: error: format not a string literal and no format arguments [-Werror=format-security]
                                        ttybarstring[current_char]);
                                        ^~~~~~~~~~~~
cc1: some warnings being treated as errors
```

This PR fixes those errors.

PR #357, opened 2020-05-28
Add slightly more strict compiler settings

hsheth2:
Follows up on #355.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -17,7 +17,7 @@ cava_CPPFLAGS = -DPACKAGE=\"$(PACKAGE)\" -DVERSION=\"$(VERSION)\" \
 cava_CFLAGS = -std=c99 -Wall -Werror -Wextra -Wno-unused-result -Wno-unknown-warning-option -Wno-maybe-uninitialized
 
 if OSX
-    cava_CFLAGS += -DNORT
+    cava_CFLAGS += -DNORT -Wno-implicit-function-declaration
 else
     cava_LDFLAGS += -lrt
 endif
--- a/Makefile.am
+++ b/Makefile.am
@@ -17,7 +17,7 @@ cava_CPPFLAGS = -DPACKAGE=\"$(PACKAGE)\" -DVERSION=\"$(VERSION)\" \
 cava_CFLAGS = -std=c99 -Wall -Werror -Wextra -Wno-unused-result -Wno-unknown-warning-option -Wno-maybe-uninitialized
 
 if OSX
-    cava_CFLAGS += -DNORT -D_XOPEN_SOURCE_EXTENDED
+    cava_CFLAGS += -DNORT
 else
     cava_LDFLAGS += -lrt
 endif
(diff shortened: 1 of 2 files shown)


hsheth2:
Looks like all the macos builds output this warning:
`warning: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [-Wunknown-warning-option]`

Adding `-Werror` turned that warning into an error, causing the macos build to fail.

hsheth2:
This looks like a macos specific issue, and I'm not sure what the right way to fix it would be. @karlstav could you take a look?
```
input/portaudio.c:29:44: error: implicit conversion from 'int' to 'int16_t' (aka 'short') changes value from 32768 to -32768 [-Werror,-Wconstant-conversion]
    int16_t silence_buffer[PORTBUFSIZE] = {SAMPLE_SILENCE};
                                          ~^~~~~~~~~~~~~~
input/portaudio.c:7:24: note: expanded from macro 'SAMPLE_SILENCE'
#define SAMPLE_SILENCE 32768
```

karlstav:
well 32768 would be -32768 when converting to int16_t so i just changed the value to it. To be honest I have no idea what this code tries to do. Changing that value to whatever changes nothing in the output.

karlstav:
oh well new error. 

```

output/terminal_bcircle.c:70:9: error: implicit declaration of function 'mvaddwstr' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        mvaddwstr(y, x, bars[7]);

```

there is no doubt that ncurses is working. some issue with the widechar support. or rather the exposing of such:

https://stackoverflow.com/questions/48042203/curses-library-doesnt-support-wide-char-on-os-x-high-sierra

probably realted

hsheth2:
I just suppressed the implicit function declarations error -- if it was really an issue it would error out at when linking.

karlstav:
I think we should try to fix it, would not `implicit declaration of function` mean that it's working despite not really being able to find that function in the included `ncurses.h` library? 

the homebrew version of ncurses is 6.2:

but i get 5.7 when running:
```
Karls-MacBook-Air:~ karl$ otool -L /usr/lib/libncurses.dylib 
/usr/lib/libncurses.dylib:
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)
```

so I think the problem is that it defaults to some ancient bundled version as discussed here:

https://github.com/hishamhm/htop/issues/301

karlstav:
I got it.

the wide functions are surrounded by `ifdef _XOPEN_SOURCE_EXTENDED` in the bundled macOS curses. This is set in top of cava.c and not included in `output/terminal_ncurses.c` setting it in `Makefile.am`solves the problem.

pull and test it again.

karlstav:
I see now that this was already what you were trying to do in the first place. I just removed it from `cava.c`and defined it in `Makefile.am` for everything. Do you think that might be a problem?

Sorry for messing up your PR, do you want to create a new one ore should I just commit your changes myself?

hsheth2:
No worries - I was trying a similar thing but couldn't get it to work quite right. Anyways, I merged in your changes but it looks like Github isn't detecting that it was updated. I'm going to close this one and open a new one so that it will pick up the changes

PR #405, opened 2021-05-17
GCC 11 compatibility

Xiretza:
GCC 11 introduced some new warnings, one of them is real, the other has to be suppressed.


karlstav:
hi @Xiretza,

thanks for the heads up on the new error added in gcc11 (or 10?). you seem to have added the parameter incorrectly, so I just introduced the fix myself.

I should also maybe just change that function...

Xiretza:
Oh well, guess I'm the only one who cares about meaningful commit messages...

PR #421, opened 2021-11-23
Repaired -lrt lib

Vojta3310:
The only + on 
cava_LDADD += -liniparser
does not solve it. So I tried to repair it, but it is my first makefile.am so sorry if it is not a good way.


Vojta3310:
You were faster.

PR #432, opened 2022-02-05
Fix SDL bar glitch

Jocker666z:
Removal of the total height between 2 rectangles of different height.

Capture d’écran du 2022-02-05 14-54-00

### DIFF
--- a/output/sdl_cava.c
+++ b/output/sdl_cava.c
@@ -80,9 +80,9 @@ int draw_sdl(int bars_count, int bar_width, int bar_spacing, int remainder, int 
             // clear bar from previous frame if lower
             if (bars[bar] < previous_frame[bar]) {
                 fillRect.x = bar * (bar_width + bar_spacing) + remainder;
-                fillRect.y = height - previous_frame[bar];
+                fillRect.y = 0;
                 fillRect.w = bar_width;
-                fillRect.h = previous_frame[bar];
+                fillRect.h = height;
                 SDL_SetRenderDrawColor(gRenderer, bg_color.R, bg_color.G, bg_color.B, 0xFF);
                 SDL_RenderFillRect(gRenderer, &fillRect);
             }


karlstav:
strange glitch, looks like it only happens on macos, some strange half pixel stuff maybe. nice catch, don't think there is any actual performance gain in not clearing the whole bar either.

Jocker666z:
> strange glitch, looks like it only happens on macos, some strange half pixel stuff maybe. nice catch, don't think there is any actual performance gain in not clearing the whole bar either.

This is not Mac OS, but debian unstable here, I also had it with a raspberry 3 and raspbian stable.

karlstav:
that's strange. I got the glitch on macos big sur with sdl version 2.0.9, but not on ubuntu 21.10 with sdl 2.0.14. what version of sdl are you running? looks like 2.0.14 on stable on 2.0.20 on unstable.

resolution? scaling?

Jocker666z:
sdl -> 2.0.20
Screenshot make with no resolution/scaling change. 

The glitch often appears when changing windows on my main machine.
On Rpi 3, sometimes appears when the processor is in overflow.

karlstav:
on my mac it happens all the time without your fix. 

cava

also background color is completely broken on my mac, seems to flicker on and off. how is that for you? try setting it to white in the config

karlstav:
looks like re-drawing the entire thing if there is a change is the way to go. can you try the latest change @Jocker666z ?

Jocker666z:
No problem here with https://github.com/karlstav/cava/commit/6e594c467a407835e73e50c9abc414f94d453ddb

I had artifacts when I resized the window, they are no longer there.

I will test on the raspberry tonight (I'm going to work in 30min), to see if it doesn't have an impact on the performance.

karlstav:
That's good to know. Thanks!

PR #438, opened 2022-03-13
Add new raw binary output for "noritake" VFD bitmap format

elafargue:
This makes it possible to use the output of cava on Noritake VFD displays directly with no additional processing.

It can  achieve framerates of 70+ fps on 3900 series displays on a Raspberry Pi with no particular effort.

Noritake bitmap raw output bar height is configurable with height up to 32 pixels.

Issue #439

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,7 +4,7 @@ ACLOCAL_AMFLAGS = -I m4
 
 bin_PROGRAMS = cava
 cava_SOURCES = cava.c config.c input/common.c input/fifo.c input/shmem.c \
-               output/terminal_noncurses.c output/raw.c
+               output/terminal_noncurses.c output/raw.c output/noritake.c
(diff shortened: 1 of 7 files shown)


karlstav:
hi @elafargue,

I have read through your patch and I think we need to do this a bit different.

just leave the raw output module as it is. Rather create a new output module called "notirake" or even better the name of the standard that these displays use (if any).

The new bar height config option is great, just leave it. I will start using it for the other output modules as well.

elafargue:
Sounds good, I'll refactor!

karlstav:
sorry for the late follow up here, this looks a lot better. But it would be best if raw.c could be left completely unchanged. I know that the way it is now is not ideal, but this is just to be sure to not introduce unwanted changes.

Also you seem to have left some extra changes in the example config file

PR #470, opened 2022-09-24
Implement different orientations of output for SDL

DziubanMaciej:
Currently cava only supports bars that are aligned to the bottom of the window and point upwards. This commit allows them to be aligned to any edge of the window.

For now it is only implemented this for SDL, because it was the simplest, but I may also do it for ncurses output.

Issue link: https://github.com/karlstav/cava/issues/471

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -552,15 +552,17 @@ as of 0.4.0 all options are specified in config file, see in '/home/username/.co
             }
 
             // checks if there is stil extra room, will use this to center
-            remainder = (dimension_bar - number_of_bars * p.bar_width - number_of_bars * p.bar_spacing +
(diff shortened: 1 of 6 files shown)


karlstav:
looks cool, can you try to make the linter happy? insecure about the macos build error.

DziubanMaciej:
Sure, I'll look into the linter.

As for macos build error, I don't have a working setup, but I think I know what's going on. The function `memset_pattern4` is defined in `string.h` only if `_DARWIN_C_SOURCE` is defined. SDL headers define it, but they are included AFTER string.h, so it has no effect. I'll try to move some includes around and cross my fingers that it helps :D

PR #489, opened 2023-03-09
Iniparser + meson 

LukashonakV:
1. Meson configs. Provide an opportunity to build cava using Meson
    1. `make`
    2. `make clean && make`
2. iniparser library in some systems is stored under separate directory. For example in Gentoo it's sits : /usr/include/iniparser4/*


karlstav:
I know the intent here is good, but do we really want to maintain two build systems? (three if we count the visual studio one). I guess if we could replace everything with meson that would be potentially something, but I'm afraid it would be  too much work for the potential benefit.

LukashonakV:
Hi @karlstav . I've got your point. Yes, agree. + I found it's possible just to add meson wrapped into meson https://github.com/mesonbuild/wrapdb
So actually this PR can be closed. The only thing still actuall there is: iniparser.h .

PR #495, opened 2023-04-13
Libcava proto

LukashonakV:
Hi @karlstav ... actually this PR provides an opportunity to build libcava as a shared library. I have checked .. and built external Waybar application using the library is built by this PR. And it works great. 
Actually my branch is named proto  - means prototype.Tomorrow I'll provide additional branch with cava application which will use exact it's own libcava

To compile: make
To install library: make install

### DIFF
--- a/example_files/test_configs/pipewire_zero_test
+++ b/example_files/test_configs/pipewire_zero_test
@@ -1,0 +1,16 @@
+## test config file for CAVA, testing if zero pulseaudio input gives zero bar height output
+
+[general]
+
+draw_and_quit = 60
+zero_test = 0
+non_zero_test = 0
+
+[input]
+method = pipewire
+source = auto
+
+[output]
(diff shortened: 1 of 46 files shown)


LukashonakV:
@karlstav, can you please create two separate branches: cavalib_proto and cava_proto. 
Actually, there are two reasons:
1. My refactoring is based on meson.  So if you approve PR to master it will break current cava. Due to these changes, it's needed to be fixed autotols configuration
2. I'm thinking of: to spread cava application from the cavalib and provide it's own meson configuration. In such case, it'll be simple to add runtime dependency during cava building. Then, meson can automatically download cavalib (if it's not presented in the system) from the separate branch and compile it, too.

Once both branches are ready,  then you can decide how to release the new chain: libcava and based on it, cava application . Please contact me for any unclear things.

PR #540, opened 2023-12-31
OSS input backend and general improved FreeBSD support

bsdcode:
- OSS: add 'enable-oss-input' configure option
- OSS: set to default input method if enabled
- sndio: add to input method selection sequence with high priority
- sndio: fix small bug in channel setting
- FreeBSD: add 'enable-cava-font' configure option, which creates vt(4) font cava.fnt from cava.psf
- doc: extended documentation in example config file and in README.md

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,7 +7,7 @@ cava_SOURCES = cava.c cavacore.c config.c input/common.c input/fifo.c input/shme
                output/terminal_noncurses.c output/raw.c output/noritake.c
 cava_CPPFLAGS = -DPACKAGE=\"$(PACKAGE)\" -DVERSION=\"$(VERSION)\" \
            -D_POSIX_SOURCE -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE_EXTENDED \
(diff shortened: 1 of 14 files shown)


karlstav:
the build on linux fails @bsdcode.

```
input/oss.c:16:49: error: ‘AFMT_S32_LE’ undeclared (first use in this function); did you mean ‘AFMT_S16_LE’?
...
```


looks like normal ubuntu also have the `sys/soundcard.h` so it tries to build the oss.c.

should it have worked? Not sure how linux relates to oss.

maybe just make the check so it only builds oss on freebsd. not sure if we can have github runners on freebsd, but would be nice in this case.

bsdcode:
I didn't know that ubuntu (or any linux distro) includes a `sys/soundcarde.h` by default. In general OSS can be built on linux, but it's AFAIK not common and it would differ anyway. I would rather limit it to the FreeBSD built as you suggest. I will update the `configure.ac`.

I have no experience with github runners, but I can have a look at it.

karlstav:
Looks like it's only linux, mac and windows.

PR #548, opened 2024-02-01
Remove double word.

lorenzsichert:


karlstav:
well spotted, however this statement is no longer true, so i will have to change it.

Issue #556, opened 2024-03-29
Keyboard controls are ignored when pausing PipeWire input


gardk:
**Describe the bug**

Title.

Found this cause I wanted cava to only visualize my music and not all desktop sounds :)

I've found that resizing the window causes the bars to actually reset, unless you pressed Q before doing so.

**To Reproduce**

1. Have MPD (only source I've tested with) running in the background
2. Run cava
3. Using qpwgraph or similar, attach cava to MPD's output instead of the default sink
4. Pause music

**Expected behavior**

Input should not be ignored and bars should return to zero.

**Screenshots**

This picture is from when the music is paused
![image](https://github.com/karlstav/cava/assets/51994294/3327405a-66f9-42c2-8660-2c6cf3b20015)

**Desktop:**

 - OS: NixOS
 - Version: 24.05 unstable

**Terminal emulator**
Tested in kitty 0.33.1 and alacritty 0.13.2

karlstav:
this might be intended behavior, not the key controls, those should work (if the sleep timer is not configured). 

When the source is paused, there is no data coming in, i would have to add a function to reset the buffer to zero after some time without new input data. I think i already have it in the fifo input.

gardk:
Yeah, the bars freezing does make sense actually, but it would be nice to have them reset to zero after some time :)

karlstav:
was finally able to reproduce. The issue is that the main thread is waiting for the audio to terminate, but the on_process callback is not being fired. Need to think of some solution.

karlstav:
I guess that was what this was for: 0d773076063dbd393dc9f7adfe78f9f5ca1dce0d will re-add it make some signal to the thread when terminating the audio.

gardk:
You can make on_process keep firing by setting `PW_KEY_NODE_ALWAYS_PROCESS`, seems to fix this issue completely.

karlstav:
it does! and then the bars also go down when pausing.

I will consider this resolved then, thanks @gardk for the contribution!

PR #557, opened 2024-03-29
fix source being ignored when using pipewire

gardk:
I found that whatever I put in the source field of [input] was always ignored by cava when using PipeWire.

This PR fixes that issue, but it also brings out another issue that is actually reproducible without this patch. Described in #556.

Haven't been able to track down why that's happening.

### DIFF
--- a/input/pipewire.c
+++ b/input/pipewire.c
@@ -85,8 +85,12 @@ void *input_pipewire(void *audiodata) {
     props = pw_properties_new(PW_KEY_MEDIA_TYPE, "Audio", PW_KEY_MEDIA_CATEGORY, "Capture",
                               PW_KEY_MEDIA_ROLE, "Music", NULL);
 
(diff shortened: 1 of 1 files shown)


karlstav:
hi @gardk,

I always found these settings a bit flimsy in pipewire. In my case now the source is ignored both with and without your commit. Maybe it is related to the fact that I am on ubuntu and stuck with pipewire 0.3.79

karlstav:
there it worked, not sure why. maybe my pipewire was in some faulty state,

gardk:
Nice, so you got it tested that it did what it says it did? I was a bit worried when you merged and it wasn't working for you :P

karlstav:
yes, I can now specify a client like spotify and pause the music to reproduce the issue. What is strange is that if i specify some random string as the source it will still use the default output. Might just be a feature in pipewire.

PR #565, opened 2024-05-21
config. getenv variables

LukashonakV:
HI @karlstav , this MR covers cases when 3-rd party applications provide configPath (see Waybar user - Alexays/Waybar#3276). Users want to keep environment variables in config. I think it's useful to extend config.c functionality to expand user environment variables. Such code tries to expand user environment variables which are started with $ character (such as $HOME, $XDG_CONFIG_HOME, etc.)

### DIFF
--- a/config.c
+++ b/config.c
@@ -469,6 +469,41 @@ bool load_config(char configPath[PATH_MAX], struct config_params *p, bool colors
         }
 
     } else { // opening specified file
+        // expand env variables in configPath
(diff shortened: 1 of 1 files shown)


karlstav:
not really sure about this one @LukashonakV, cava is a terminal application meant to be run in a shell and environment variables will already be expanded by the shell. If cava is wrapped into something else, then the expansion of the variables should probably be done elsewhere.

PR #566, opened 2024-05-28
Fix typo on cava.c

Polkorny:
The patch is currently applied to the cava-alsa package on debian


karlstav:
Quitting is with two t's!? Who could have guessed.

Thanks

PR #569, opened 2024-05-30
Fix crash when using pulseaudio

mohd-akram:
The main loop should be freed only after it has stopped running.

I get a segmentation fault on macOS without this patch.


karlstav:
looks good

PR #576, opened 2024-07-18
Synchronized Sync (Mode 2026)

ThatOneCalculator:
Deprecates the Alacritty synchronization option for Mode 2026 (https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ missing
 version
 config_file.h
 .*
+cava
--- a/cava.c
+++ b/cava.c
@@ -1098,10 +1098,9 @@ as of 0.4.0 all options are specified in config file, see in '/home/username/.co
 
 // output: draw processed input
 #ifdef NDEBUG
-                if (p.sync_updates) {
-                    printf("\033Ptmux;\033\033[?2026h");
-                    fflush(stdout);
-                }
+                printf("\033P[2026;2$y");
+                fflush(stdout);
+
                 int rc;
 #ifdef _MSC_VER
                 QueryPerformanceCounter(&t1);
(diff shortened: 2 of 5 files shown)


ThatOneCalculator:
@karlstav Tested in ghostty, alacritty, kitty, and gnome console. This is a naive implementation but it does seem to work.

karlstav:
hi @ThatOneCalculator,

sorry for the late feedback. There is one thing I don't understand. Why is there both a begin and end update before and after the draw function? Shouldn't there be BSU (2026h) before (line 1101) and ESU (2026l) after (line 1152)?

ThatOneCalculator:
To be honest, this was a very naive implementation -- I did it to the point where it worked, but I think that I probably made a mistake or two along the way. I am not a C programmer nor a terminal UI programmer, but this fixed works for me, although I doubt it's the right way to do things. Something about having the update go through even though it immediately gets ended makes ghostty not tear. I think it could probably work as you described having the BSU before and the ESU after. Feel free to change my implementation!

karlstav:
If it works, it works. A couple of things though: if the old alacritty sync thing has not been deprecated by alacritty i think we should keep it along side this one. Also I think this sync method also must be a config option and that it must be default off. Never know if it can cause issues.

ThatOneCalculator:
Afaik alacritty now uses the new standard. Tested this on kitty/ghostty/alactitty/foot and all worked

karlstav:
noted, https://github.com/alacritty/alacritty/pull/7365

just re apply the config option (I still believe it should be optional) and squash the commits and then I can merge

ThatOneCalculator:
Should hopefully be able to do soon :D

ThatOneCalculator:
Oh shit I kinda forgot about this

ThatOneCalculator:
@karlstav there we go! Sorry about the very long delay

ThatOneCalculator:
Tested properly now. Works completely as expected!

karlstav:
no problem, hmm looks like I can squash the commits when i merge, i did not know that. seemed excessive with 10 commits for the 5 lines of code change.

PR #610, opened 2024-12-29
Fix unlinked issue in README

arithefirst:
While skimming through the README, I noticed the following segment

> ### Gradient not working in Konsole
> Konsole simply does not support this. #​194

I tried to click on the issue, but then remembered that GitHub doesn't support directly linking Issues/PRs in markdown files, only from other Issues of PRs. This PR just adds a link to #​194 
on the text, for user convince (See below).

# Before:
image

# After:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -616,7 +616,7 @@ Actually, `setfont` is supposed to return the default font, but this usually isn
 
 ### Gradient not working in Konsole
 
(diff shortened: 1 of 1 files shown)


arithefirst:
Ignore the force push, I just rebased the commit name to have better grammar.

PR #614, opened 2025-01-27
Emulate the "line style" spectrum analyzer from Winamp 2

pmarks-net:
Here's a comparison video (no sound) between Winamp 2.91 and this shader:

[sandstorm-trim.webm](https://github.com/user-attachments/assets/3381e04f-9fa6-467a-9524-8bdea2e0755d)


karlstav:
awesome! I will add it to the config.c file so it gets embedded at compile time.

the system for embedding and installing the shaders is a bit complicated at the moment.

PR #624, opened 2025-02-23
CI: Add Windows build

Biswa96:


karlstav:
can you rename them so they are build-windows not just build-Debug? maybe also debug build is enough?

Biswa96:
Done.

PR #627, opened 2025-02-27
Add support for mingw toolchain in Windows

Biswa96:


Biswa96:
tl;dr The changes are not due to mingw toolchain. These fix the assumptions and mistakes with Microsoft Visual Studio compiler.

karlstav:
nice! yes the windows support was thrown together sitting in visual studio, didn't really expect it to catch on. Maybe the cmake could be expanded to be used in visual studio as well and get rid of some of the hardcoded project files. Not sure how to do the nuget stuff though.

Biswa96:
With these changes, cava is now in msys2/mingw repository https://packages.msys2.org/base/mingw-w64-cava

PR #637, opened 2025-05-06
Orientation cycling

globbertot:
cava supports orientations but are only configurable in the config file, this allows for runtime change using the 'o' key

issue: #638


karlstav:
cool, not sure how useful, but I will merge it if you can fix the linter issue. Would be more useful if key command changes could be saved to the config file, but that seems to complicated to be worth it..

globbertot:
Oh my bad, sorry about that. I might work on saving the orientation as it sounds interesting as a option

karlstav:
saving is not straight forward at all with ini configs on linux I'm afraid

Issue #645, opened 2025-06-08
Crash when building and running master branch on windows


rohanb1216:
**Describe the bug**

Crash when building and running master branch on windows, event viewer shows Kerne32,dll crash.

**To Reproduce**

1. Pull latest commit
2. Compile with either Build/Release configs
3. Run

**Expected behavior**

Cava should run and show output.

**Desktop (please complete the following information):**

Windows 11 24H2
**Terminal emulator**
pwsh 7.5.1

PR #646, opened 2025-06-08
allocate memory for theme in load_config function

rohanb1216:
#645


karlstav:
Well spotted, I mainly work on linux so testing on windows is limited.

Maybe I should just use iniparser on windows as well. Not sure if it's even possible though.

Issue #651, opened 2025-06-30
Handle changing default device on Windows


rohanb1216:
**Is your feature request related to a problem? Please describe.**
On windows, audio capture fails when the default device changes.

**Describe the solution you'd like**
Audio monitoring should switch to using the new device instead of continuing to use the old one.

**Additional context**
This is just an unhandled/unported feature from @quantum5's implementation used in cava.

karlstav:
squashed and merged, nice work!

PR #652, opened 2025-06-30
Handle default device change on windows

rohanb1216:
Implements @quantum5 's notification device client in C to monitor for default device changes.
Handles https://github.com/karlstav/cava/issues/651


karlstav:
nice!

PR #656, opened 2025-07-15
[docs] add BlackHole as macOS output source

cybardev:
I have an Apple Silicon MacBook Air M1 (2020).

- Soundflower doesn't work at all.
- Background Music works as good as a broken clock telling the time right twice a day.
- BlackHole works flawlessly.

Thought it should be documented for the benefit of macOS users who want to enjoy `cava` and retain sanity (trying to get the other two to work has been chipping away at mine for ~2 years).

### DIFF
--- a/README.md
+++ b/README.md
@@ -545,6 +545,14 @@ Install [Background Music](https://github.com/kyleneideck/BackgroundMusic) which
 source = "Background Music"
 ```
 
+**BlackHole**
+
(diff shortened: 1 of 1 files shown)


cybardev:
Loosely related:

https://github.com/karlstav/cava/blob/cc1579526423be50821961f5a23bfeb93ae14f8c/README.md?plain=1#L555-L556

This line under Soundflower should be removed, since we already have it here:

https://github.com/karlstav/cava/blob/cc1579526423be50821961f5a23bfeb93ae14f8c/README.md?plain=1#L533-L538

We don't have it for Background Music, and I didn't add it for BlackHole. Should be consistent for all 3 imo.

karlstav:
If this is the defacto standard way of doing this kind of thing in macos now, I am all for moving it to the top recommended method of capturing audio.

Maybe also just remove soundflower?

cybardev:
I'm not sure honestly. The Soundflower repo says a new version is in the works, but I would rather bet on horses than unicorns (something that exists and works vs not). Maybe we can remove it for now and add back if there is a new version later.

And sure, if you want you can move BlackHole above Background Music. I just added it in alphabetical order (Ba < Bl). I don't really know if BlackHole is the best way, but I have been trying Background Music on and off the last few years and it "works" for a bit and then crashes. If I have to keep manually restarting the program, it's not really working as it should. BlackHole just adds devices to the system settings somehow. Doesn't even need to keep running in the background (at least not that I can see), and hasn't randomly stopped working (yet). So... it's the best _for me_ but not sure how it is in general for others.

PR #668, opened 2025-07-27
generate downmixed stereo buffer for Windows

theeternalsw0rd:
Always feed a buffer of at most 2 channels to cava regardless of audio output to device. There may still be edge cases that need covered where it has the comment // Unsupported format, handle error (line 405). This handles 16-bit pcm multichannel, 32-bit pcm multichannel, and 32-bit floating point multichannel.

### DIFF
--- a/input/winscap.c
+++ b/input/winscap.c
@@ -10,10 +10,130 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <windows.h>
+#include <stdint.h>
+#include <stddef.h>
 
 #include <functiondiscoverykeys.h>
 
 #include "common.h"
+
+static void make_stereo_waveformatex(const WAVEFORMATEX *orig, WAVEFORMATEX *stereo) {
+    *stereo = *orig;
+    stereo->wFormatTag = WAVE_FORMAT_PCM;
+    stereo->nChannels = 2;
+    stereo->wBitsPerSample = 16;
+    stereo->nBlockAlign = stereo->nChannels * (stereo->wBitsPerSample / 8);
+    stereo->nAvgBytesPerSec = stereo->nSamplesPerSec * stereo->nBlockAlign;
+    stereo->cbSize = 0;
+}
+
+// Downmix interleaved 16-bit PCM to stereo
+static void downmix_to_stereo_s16(const int16_t *src, int16_t *dst, size_t frames, int channels) {
+    size_t i;
+    for (i = 0; i < frames; ++i) {
+        int left = 0, right = 0;
+        if (channels >= 2) {
+            left += src[i * channels + 0];
+            right += src[i * channels + 1];
+        }
+        if (channels >= 3) {
+            left += (int)(src[i * channels + 2] * 0.7f);
+            right += (int)(src[i * channels + 2] * 0.7f);
+        }
+        if (channels >= 5) {
+            left += (int)(src[i * channels + 4] * 0.7f);
+            right += (int)(src[i * channels + 5] * 0.7f);
+        }
+        if (channels >= 7) {
+            left += (int)(src[i * channels + 6] * 0.7f);
+            right += (int)(src[i * channels + 7] * 0.7f);
+        }
+        if (left > 32767)
+            left = 32767;
+        if (left < -32768)
+            left = -32768;
+        if (right > 32767)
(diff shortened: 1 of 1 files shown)


theeternalsw0rd:
After checking on my second computer, I saw that 24-bit seems to be common as well, so I added downmixing for that format.

In further testing there was inconsistency if cava was not fed 16-bit, so I added conversion for mono and stereo 24 and 32-bit as well.

Another thing I did learn is that Dolby Atmos, DTS:X, basically anything with Spatial processing enabled is not available via loopback. So getting that to work would either require developing an audio driver to ship with cava or utilizing third-party software to handle audio mixing and virtual cables.

As it stands right now, cava freezes if fed Dolby Atmos. It seems to handle a failure properly and print that the device is not supported but then locks up. I'll see if I can determine more at a later time.

karlstav:
hi @theeternalsw0rd.

A lot of stuff going on here I can see! I'm afraid I don't have time to go through it in detail, but browsing through it looks like you have added a lot of converting that I don't understand. cava is supposed to handle s8, s16, s24 and float both mono and stereo. But you have added stuff like "convert_mono_f32_to_s16". Maybe the mono processing in cava is not working?

If cava is not working when not fed 16 bit stereo, than that needs to be fixed in cava.

down/up mixing:

https://github.com/karlstav/cava/blob/master/input/common.c#L21

stereo/mono handling is happening around here:

https://github.com/karlstav/cava/blob/master/cavacore.c#L362

theeternalsw0rd:
I've pulled the conversion of stereo and mono out. Turns out the issue my previous refactor resolved that had to do with the graph freezing before reaching zero level when going from audible audio to silence, also resolved the issue with non-16-bit audio. Quite bizarre.

input/winscap.c
```diff
@@ -132,6 +304,44 @@ struct {
...
+void write_silent_frame(struct audio_data *audio, IAudioCaptureClient *pCapture,
+                        UINT32 numFramesAvailable, UINT32 *packetLength) {
+	// Send one silent frame to the spectrometer
+	int silent_channels = audio->channels;
+	int silent_bytes = silent_channels * sizeof(int16_t); // 16-bit PCM
+	int16_t silence[2] = {0};
+
+	write_to_cava_input_buffers(silent_channels, (unsigned char *)silence, audio);
+	pCapture->lpVtbl->ReleaseBuffer(pCapture, numFramesAvailable);
+	pCapture->lpVtbl->GetNextPacketSize(pCapture, &packetLength);
```
karlstav, line 316: is it maybe just `packetLength` here, without the `&`?

karlstav: or maybe better remove the * from the write_silent_frame packetLength function argument and remove the & when calling the function.

theeternalsw0rd:
That’s probably it. I think I already corrected that in another method using the former solution. I’ll make sure it’s consistent using the latter.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

theeternalsw0rd:
20-bit audio is also used it seems. I've set up a separate branch for that since I cannot test it currently. I'll try to locate some hardware to test on. I think the virtual cable I have at work supports it, so I can check that tomorrow. The changes for the use of packetLength have been pushed.

theeternalsw0rd:
I did find a device that supports 20-bit audio, but that format is not selectable in Windows 11 sound settings. Windows itself may have removed support completely or 20-bit is only available via direct hardware or the audio driver isn't properly sending supported formats to Windows. I will not worry about it at this point. It's available on my repo under the 20-bit branch if anyone should have need. I am assuming anything with 20-bit support also supports other formats too.

karlstav:
just use the .clang-format file to make the linter happy. don't worry about the commits, I will squash them when i merge

theeternalsw0rd:
Hopefully the linter is happy now. I made the handling of overflowing lines as consistent as possible across the board too.

theeternalsw0rd:
Sorry, didn't realize the linter failure actually generates a patch that it would be happy with. I'll get that applied. I guess the reason for the inconsistent long line formatting is the linter itself being at fault.

theeternalsw0rd:
The linter supplied patch has been applied. Hopefully now it should have no reason to complain.

Issue #661, opened 2025-07-19
Pipewire linking issues and a solution


pallaswept:
First off apologies for not following the bug template but I'm sure you will see that I'm not taking short cuts ;)
I seem to have found the cause and a fix for [this issue](https://github.com/karlstav/cava/issues/422#issuecomment-994270910), and several others which had not been reported yet, but it needs some special handling to do it right. There is a TL;DR at the bottom. Following is some history and references.

cava's `source` setting, is used to fill the `target.object` property of cava's pipewire node, informing the session manager of where to link it.

#557  removed the erroneous setting of the `target.object` property to the string "auto". Because "auto" was not a valid `node.name`, the search for the node failed, the fallback was used, `stream.capture.sink = true` is not used to search for a fallback, so it used the highest priority device which had output ports and that's the default input device.

Ironically, manually selecting that same fallback input device (the mic) as cava's source, would have failed, because `stream.capture.sink = true` was set, and those nodes do not have input ports so they do not have monitor ports. Likewise, if the same default output device (speakers) were selected manually rather than using 'auto' to fall back to it, because `stream.capture.sink = true` was set, it would have worked fine, rather than falling back to the default input device.

#557  also set `stream.capture.sink = true` only for the default device when specified with "auto". This made cava work with the "auto" source (`target.object` unset), and coincidentally, because `stream.capture.sink = true` was unset for them, also made cava work with any input device nodes or filter playback nodes. These would not function otherwise because they lack the monitor ports which are needed for `stream.capture.sink = true`. 

However, lacking the `stream.capture.sink = true` property to indicate to link to monitor ports, made cava not work with the default device set manually, any non-default output device, or filter capture nodes.

https://github.com/karlstav/cava/issues/422#issuecomment-994270910 Was close but not quite correct, which will be why it didn't work for everybody. Using the `node.serial` of the node, will work, even without `stream.capture.sink = true`. Using the `node.id` is not a real thing and does not work, we can only use the `name` or `serial`, however the `node.id` and `node.serial` will often correspond, in which case that suggestion of using the `node.id` will 'coincidentally' work.

[...]

PR #670, opened 2025-08-02
Pipewire changes via #661

pallaswept:
Hi @karlstav 

I have altered this to make the new features opt-in as suggested and am just filing this PR for your consideration whenever it's a good time for you.

There are two parts I am presently confident are working but I'm unsure they are optimal - 
 - Requested latency 
   * This has increased from 10 to 10.6ms, and is then rounded up to a power of 2. It appears to work as expected, but it might be worth checking `n_samples` especially when samplerate conversion occurs (eg 44100 cava on 48k pipewire graph)
   * I have a suspicion that this may work well/better, just set to '512/48000', _without rounding up to a power of 2_, regardless of the sample rate of cava. The numerator there doesn't actually specify a sample rate, it's just a divider on a time interval - so with '512/48000' we are literally saying "give me 10.6ms intervals", even if the samplerate is not 48000Hz. Likewise, the power of 2 is only needed to avoid the graph latency being rounded down, and '512/48000' should have that effect all by itself as the required power of 2 is included.
     - One example I'm less sure of - if the graph is at 44k1, and cava is at 44k1, and cava requests latency of '512/48000' that will be 470/44100 which is less than 512/44100 which the graph would be using, so it'd have to boost the graph to 256/44100 to service it in time. So, maybe we really do need to round up. This example is the reason I went this way first.
 - Clearing the bars to zero when the node is idle
   * I just created a var initialised to zero and read that in. It works great, but it's not as elegant as what you did in the FIFO input.
   * That fill with zeros occurs at 60FPS, I chose this as that's cava's default and the framerate config isn't exposed to the inputs. Perhaps it'd be better to just fill the entire buffer in one go (I tried this but it wasn't cava's usual pretty 'fade out' just sudden emptiness), or alter the framerate somehow?

All that being said, I'm now at 24 hours of constant audio playing through cava and zero xruns, which is very nice :) I hope this is useful to you!

### DIFF
--- a/input/pipewire.c
+++ b/input/pipewire.c
@@ -202,19 +202,19 @@ void *input_pipewire(void *audiodata) {
         pw_properties_set(props, "channelmix.upmix", "true");
 
         if (data.cava_audio->channels < 2) {
-        // N to 1 with all channels shown
-        params[0] = spa_format_audio_raw_build(
-            &b, SPA_PARAM_EnumFormat,
-            &SPA_AUDIO_INFO_RAW_INIT(.format = audio_format, .rate = data.cava_audio->rate,
-                                     .channels = data.cava_audio->channels, ));
+            // N to 1 with all channels shown
+            params[0] = spa_format_audio_raw_build(
+                &b, SPA_PARAM_EnumFormat,
+                &SPA_AUDIO_INFO_RAW_INIT(.format = audio_format, .rate = data.cava_audio->rate,
+                                         .channels = data.cava_audio->channels, ));
         } else {
-        // N to 2 with all channels shown
-        params[0] = spa_format_audio_raw_build(
-            &b, SPA_PARAM_EnumFormat,
-            &SPA_AUDIO_INFO_RAW_INIT(.format = audio_format, .rate = data.cava_audio->rate,
-                                     .channels = data.cava_audio->channels,
-                                     .position = {SPA_AUDIO_CHANNEL_FL,
-                                                  SPA_AUDIO_CHANNEL_FR}, ));
+            // N to 2 with all channels shown
+            params[0] = spa_format_audio_raw_build(
+                &b, SPA_PARAM_EnumFormat,
+                &SPA_AUDIO_INFO_RAW_INIT(.format = audio_format, .rate = data.cava_audio->rate,
+                                         .channels = data.cava_audio->channels,
+                                         .position = {SPA_AUDIO_CHANNEL_FL,
+                                                      SPA_AUDIO_CHANNEL_FR}, ));
         }
     } else {
         // N to 2 with only FL and FR shown
@@ -222,7 +222,6 @@ void *input_pipewire(void *audiodata) {
             &b, SPA_PARAM_EnumFormat,
             &SPA_AUDIO_INFO_RAW_INIT(.format = audio_format, .rate = data.cava_audio->rate,
                                      .channels = data.cava_audio->channels, ));
-
     }
 
     data.stream = pw_stream_new_simple(pw_main_loop_get_loop(data.loop), "cava", props,


karlstav:
> * That fill with zeros occurs at 60FPS, I chose this as that's cava's default and the framerate config isn't exposed to the inputs. Perhaps it'd be better to just fill the entire buffer in one go (I tried this but it wasn't cava's usual pretty 'fade out' just sudden emptiness), or alter the framerate somehow?

cava reads audio and prints asynchronously, so using the term FPS here doesn't really make any sense. As the "default" for cava is to read 512 samples per "execution", then using a timeout of 10 ms should be fine (512 samples /44100 samples per second ≈ 10ms), 

Using the `reset_output_buffers` function as the fifo input does, would maybe make most sense (it is in the common.h). There should be a certain fade as there is some smoothing filters in place, but certainly a bit faster than cutting the audio in "always process mode" (because the cava buffers is longer than 512 samples)

karlstav:
this:
```
    pw_properties_set(props, PW_KEY_STREAM_MONITOR , "true");
```
causes some serious harm to the signal. Just tested it now, the frequency is doubled and a lot of noise is introduced. tried to read the docs, but not a lot of details:

"might select a less accurate but faster conversion algorithm."

maybe it is meant for just checking whether or not there is any sound from the input.

karlstav:
the debugger says 1024 now, but that is of course stereo:

image


so that is just fine.

pallaswept:
> the debugger says 1024 now, but that is of course stereo:
> so that is just fine.

:tada: That was the part I was most unsure about. Ironically everything else went wrong :D
I am pretty sure we could get that down to 470(940 in stereo) by just requesting 512/48000. That would obviously be 512 at 48k, when the daemon is at 48k. But if the daemon is at 44k1, we have just requested 512/48000 == 470/44100 and the daemon will be at a power of 2, 512/44100, so it will have to come down to a quantum of 256/44100 = 5.8ms in order to serve cava the 470 it requested.

So basically, the way it is now avoids any 'our 10ms became 5.something' situations, but it does have that small cost of 'our 10ms became 11.6ms' , when cava is at 44100 and the daemon is at 48000. 
Probably for pipewire users, it's best to set cava to 48k, unless they play almost exclusively 44k1 media it's likely that the graph is running at 48k, so it will avoid sample rate conversion in the daemon, as well as this timing thing. Perhaps I should mention that in the examples config for pipewire?

> causes some serious harm to the signal. Just tested it now, the frequency is doubled and a lot of noise is introduced. tried to read the docs, but not a lot of details

Sorry! I pored over those docs for ages and I was on the fence but obviously made the wrong call, will trash that. 

> cava reads audio and prints asynchronously, so using the term FPS here doesn't really make any sense. 

I do know what you mean, I felt weird typing it - but the thing is, the rate at which I write to the buffers in that state, determines the actual visible framerate of the updates from cava. 

If I feed it zero-filled buffers either too small or too slowly, the bars actually grow, it's quite odd :D I honestly think that what's happening is that I'm triggering the overflow protection [here](https://github.com/pallaswept/cava/blob/b0cd89e79ea48869f9ad49c8386f4cbc0bc659b4/input/common.c#L12) which does the same thing as `reset_output_buffers()` but also increments `samples_counter` at the end.

> Using the `reset_output_buffers` function as the fifo input does,

I originally tried to use `reset_output_buffers(data->cava_audio);` and it doesn't update the bars at all, which is why it's currently using `write_to_cava_input_buffers()` with a bogus buffer in that same place.
I figured that you'd know why `reset_output_buffers()` wasn't working in that same place. Is it maybe because that function doesn't increment `samples_counter`? I'm looking at these guys:
https://github.com/pallaswept/cava/blob/b0cd89e79ea48869f9ad49c8386f4cbc0bc659b4/cava.c#L1097
https://github.com/pallaswept/cava/blob/b0cd89e79ea48869f9ad49c8386f4cbc0bc659b4/cavacore.c#L352

...but fifo.c doesn't change that so that's where I got really confused.

karlstav:
I am confused by the usage 48000 sample rate here. Isn't the "default" 44100. cava also defaults to using this. But if pipewire is most likely to default to 48k we can just default to that as well, no need to bother the user with recommendations to change the config.

I think you are correct about `reset_output_buffers()`, it should set the `samples_counter`to `audio.cava_buffer_size`. To make sure all internal buffers are cleared. You can test it, just add a commit if it works:

```
void reset_output_buffers(struct audio_data *data) {
    struct audio_data *audio = (struct audio_data *)data;
    pthread_mutex_lock(&audio->lock);
    for (uint16_t n = 0; n < audio->cava_buffer_size; n++) {
        audio->cava_in[n] = 0;
    }
    audio->samples_counter = audio->cava_buffer_size;
    pthread_mutex_unlock(&audio->lock);
}
```

The growing bars is most likely the ausosens functionality kicking in, because there is still something in the buffers. Did we actually check that the fifo input is working correctly, I suspect it will have the same bug.

pallaswept:
> Isn't the "default" 44100

Pipewire is all new-fangled and uses 48000. 
https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/pipewire/settings.c#L23

I think I would need to make a small change to config.c, to change to 48k by default only for pipewire (changing it for everything sounds scary?) but if that's OK with you I think that should work no problem.

> You can test it, just add a commit if it works:

:tada: that works perfectly! Stops exactly in line with the falloff of the bars from the smoothing, consistent FPS, everything normal and pretty. Nice one!

> Did we actually check that the fifo input is working correctly, I suspect it will have the same bug.

I haven't but I'll test that it works with that new function (and presumably doesn't work without) before I push them all. I

karlstav:
> I think I would need to make a small change to config.c, to change to 48k by default only for pipewire (changing it for everything sounds scary?) but if that's OK with you I think that should work no problem.

sounds good. mention that the default for pipewire is 48k it in the config too, somewhere around:

https://github.com/karlstav/cava/blob/master/example_files/config#L115



> 🎉 that works perfectly! Stops exactly in line with the falloff of the bars from the smoothing, consistent FPS, everything normal and pretty. Nice one!

actually a bit surprising, just found a bug where `audio.cava_buffer_size` would be set too small compared to what is actually used. just pushed a fix for it:

https://github.com/karlstav/cava/commit/14bf56e6b71c0e3642d67da9eede7d3c8d0213eb

maybe it would still work ok in most cases.

pallaswept:
All taken care of, and formatted for the linter so it passes CI now (sorry I wasn't getting emails about it so I didn't notice my warning was CI's error)

> just found a bug where audio.cava_buffer_size would be set too small compared to what is actually used

This explains a lot! I'm pretty sure my 6-channel nodes were having real trouble with this. Makes sense, given they use triple the buffer, if the buffer is already a little small, things might not go well :) Good find!!

I think it might be all done so far? It runs really well here.

pallaswept:
Thanks heaps for all your help Karl.

karlstav:
Thank you for great contributions!

Issue #659, opened 2025-07-18
Equalizer settings aren't being read on Windows


Giovani1906:
As said in the title, it doesn't seem cava reads the equalizer values at all.
Even changed the settings to the following with no change in behaviour.
```
[eq]
1=1
2=0
3=1
4=0
5=1
```
Tried on Windows 10 and 11 with the Terminal emulator and `sdl_glsl` output method with no result on both the latest release and latest master commit.

karlstav:
The equaliser is not supported on windows at this time. Sorry about that.

I could not find [getsecnkeys](http://people.ee.ethz.ch/~bonaccos/debian/iniparser/api/iniparser_8h.html#a3e50b16a71a4fbd2183bec0424699932) for windows ini reader. But it should be able to implement without it. I'll put it on the todolist.

PR #679, opened 2025-09-03
Read eq values from config in Windows

just-koh:
Utilizes GetPrivateProfileString() to read eq values until the next key is not found. 
Satisfies #659


karlstav:
nice solution!

looks like it should be char *endptr not int.

and sorry about the strict linter.

just-koh:
No problem, glad to contribute to this project!

Issue #687, opened 2025-09-16
PipeWire input hangs indefinitely without running server


mia-0:
**Describe the bug**

The PipeWire input does not check the return value of `pw_stream_connect`. Thus, when there is no running PipeWire daemon, cava will hang indefinitely.

**To Reproduce**

1. Make sure PipeWire is not running. On most systems it uses systemd socket activation, so make sure to stop the socket unit too.
2. Run cava with the test config in `example_files/test_configs/pipewire_zero_test`

**Expected behavior**

Exit with error status.

**Desktop (please complete the following information):**

 - OS: openSUSE Tumbleweed
 - Version: 20250911

**Additional context**

Suggested simple fix here: https://github.com/karlstav/cava/compare/master...mia-0:cava:pw-fix?expand=1

I’m willing to work on this input to clean up the code and make it handle some further edge cases.

karlstav:
Looks great, just make a pull request and I can merge it right in

PR #688, opened 2025-09-16
input/pipewire: Handle stream connection failure

mia-0:
Without a runing PipeWire daemon, cava would hang indefinitely.


karlstav:
fixes #687

PR #698, opened 2025-10-12
config_clean method

LukashonakV:
Dear @karlstav, during upstream synchro into cavalib I found a problem when cava is used as a shared library. It's not a critical, but anyway valgrind complains when C++ application uses cavalib is written in C and class destructor is called. The problem not all memory allocated for the config_params is freed when class go out of the scope.
Current PR is a try to fix it. At leas is solved the issue I'm talking about.
Thanks.

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -1058,6 +1058,7 @@ Keys:\n\
                     resizeTerminal = true;
                     reload_colors = 0;
                     free(themeFile);
+                    break;
                 }
 
 #ifndef NDEBUG


karlstav:
segfaults on reload

asan gives me this:

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==324326==ERROR: AddressSanitizer: SEGV on unknown address (pc 0x7cdeb8043e0a bp 0x7ffe215fca40 sp 0x7ffe215fca00 T0)
==324326==The signal is caused by a READ memory access.
==324326==Hint: this fault was caused by a dereference of a high value address (see register values below).  Disassemble the provided pc to learn which register was used.
    #0 0x7cdeb8043e0a in bool __sanitizer::atomic_compare_exchange_strong<__sanitizer::atomic_uint8_t>(__sanitizer::atomic_uint8_t volatile*, __sanitizer::atomic_uint8_t::Type*, __sanitizer::atomic_uint8_t::Type, __sanitizer::memory_order) ../../../../src/libsanitizer/sanitizer_common/sanitizer_atomic_clang.h:84
    #1 0x7cdeb8043e0a in __asan::Allocator::AtomicallySetQuarantineFlagIfAllocated(__asan::AsanChunk*, void*, __sanitizer::BufferedStackTrace*) ../../../../src/libsanitizer/asan/asan_allocator.cpp:668
    #2 0x7cdeb8043e0a in __asan::Allocator::Deallocate(void*, unsigned long, unsigned long, __sanitizer::BufferedStackTrace*, __asan::AllocType) ../../../../src/libsanitizer/asan/asan_allocator.cpp:732
    #3 0x7cdeb8121266 in free ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:52
    #4 0x5afc138e7bf4 in config_clean /home/karl/memfixcava/config.c:1183
    #5 0x5afc138e9e9f in load_config /home/karl/memfixcava/config.c:523
    #6 0x5afc138dcbef in main /home/karl/memfixcava/cava.c:327
    #7 0x7cdeb722a577 in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #8 0x7cdeb722a63a in __libc_start_main_impl ../csu/libc-start.c:360
    #9 0x5afc138e1d14 in _start (/home/karl/memfixcava/cava+0x18d14) (BuildId: 784ac1e71c1d4aebe04c1176b68209225e73cd3c)

==324326==Register values:
rax = 0x0000000000000002  rbx = 0x00007cdeb81c5f00  rcx = 0x0000000000000000  rdx = 0x0000000000000003  
rdi = 0xbebebebebebebebe  rsi = 0xbebebebebebebebe  rbp = 0x00007ffe215fca40  rsp = 0x00007ffe215fca00  
 r8 = 0x00007ffe215fca60   r9 = 0x0000000000000001  r10 = 0x0000000000000001  r11 = 0x00005afc138e7bf5  
r12 = 0x00007ffe215fca60  r13 = 0xbebebebebebebebe  r14 = 0xbebebebebebebeae  r15 = 0x0000000000000001  
AddressSanitizer can not provide additional info.

```
this keeps going above 8:

image

but it is hard coded to use 0 -7 here:

https://github.com/karlstav/cava/blob/master/config.c#L750

so i think `i < 8` should be fine

karlstav:
also some double frees and reload only colors must be handled, this is maybe it, needs some more testing:

[fix_double_free_color.patch](https://github.com/user-attachments/files/22873987/fix_double_free_color.patch)

LukashonakV:
That's interesting.  Thank you. Let me check again.

karlstav:
overlooked some stuff! 

some of these bugs where already in the code from before.

[fix_double_free_color2.patch](https://github.com/user-attachments/files/22874046/fix_double_free_color2.patch)

karlstav:
needs some thorough testing. I will create a new branch tomorrow and we can merge your fixes and mine into that.

LukashonakV:
Hi @karlstav which OS and compiler do you use to compile application ? 
I'm trying to reproduce this in Gentoo but for some reason this works smoothy. Even when I'm forcing the loop to get into config reloading valgrind doesn't report me any memory leaks or segfaults.

Upd. Please ignore. I did wrong dev. test

LukashonakV:
Dear @karlstav  , finally this PR is ready.
Please consider to merge. What changes are;
1. virtual variable renamed to virtual_ . Reason: in C++ it's a reserved word so it does disallow the possibility to compile cava as a library when an external application is written in C++.
2. Some functionalities were moved under separate supporting functions
3. More accurate management of the allocated memory
4. On the right picture - memory leaks before. We have some. On the left - 0 memory leaks after this PR
ps_2025-10-16-01_03_55

karlstav:
Looks promising! I hope to get some thorough testing done later today.

BTW, I really dislike adding underscores to variables to fix naming collisions, it makes the code difficult to read. Can we call it "virtual_node" instead? I think it's a better name for it.

LukashonakV:
Sure.  Let me fix it 

----
Done

----
Two notes here:

1. There is a bug is presented before my PR and can be reproduced: using noncurses terminal and empty gradients in the config - to start cava. Then go to config and provide gradient configuration. Then go back to cava and press 'c' to reload colors. cava fails with the memory access
2. There it is still presented memory leaking using sdl_glsl output. Not major. But needs to be investigated further

karlstav:
this breaks the themefile thing. If there is a themefile configured then load_colors must use that instead

config.c
```diff
@@ -787,84 +761,34 @@ bool load_config(char configPath[PATH_MAX], struct config_params *p, bool colors
...
-    p->horizontal_gradient_count = 0;
-    for (int i = 0; i < 7; ++i) {
-        if (strcmp(p->horizontal_gradient_colors[i], "not_set") != 0)
-            p->horizontal_gradient_count++;
-        else
-            break;
-    }
-
     free(themeFile);
+    bool resultColor = load_colors(configPath, p, error);
```
karlstav: here there can be a:
```
bool resultColor;
    if (strcmp(p->theme, "none") != 0) {
        resultColor = load_colors(themeFile, p, error);
    } else {
    resultColor = load_colors(configPath, p, error);
}
```
or something

karlstav: p->color = strdup(iniparser_getstring(ini, "color:foreground", "default"));
    p->bcolor = strdup(iniparser_getstring(ini, "color:background", "default"));
    
    should also be in the load_colors function

LukashonakV:
Dear @karlstav , might possible I'm already annoying with this request )) Can you please check once again )
I fixed your last comment.
So the logic is split in more simple things, like to have a method which is providing theme file. Then it can be used to setup necessary colors from it. This provides more visibility in the codebase. In the same time no memory leaks is registered during test, besides these two things is mentioned above
ps_2025-10-18-15_29_21

karlstav:
looks pretty solid yea, nice work @LukashonakV!

I will do some more testing on my side before merging.

karlstav:
there is a double free of the eq arrary it needs to be removed from here:

```
diff --git a/cava.c b/cava.c
index 5975568..75a36f1 100644
--- a/cava.c
+++ b/cava.c
@@ -1441,9 +1441,6 @@ Keys:\n\
         pthread_mutex_unlock(&audio.lock);
         pthread_join(p_thread, NULL);
 
-        if (p.userEQ_enabled)
-            free(p.userEQ);
-
         free(audio.source);
         free(audio.cava_in);
         cleanup();
```

haven't been able to spot anything else yet.

LukashonakV:
Hi @karlstav , thank you. This double free I missed somewhere in the middle on the process  ) It's fixed now
Upd: there is one more commit which fixes issue number one from the list above
> There is a bug is presented before my PR and can be reproduced: using noncurses terminal and empty gradients in the config - to start cava. Then go to config and provide gradient configuration. Then go back to cava and press 'c' to reload colors. cava fails with the memory access
----
> There it is still presented memory leaking using sdl_glsl output. Not major. But needs to be investigated further

this one looks like happens somewhere deep in SDL library... not in the cava code

Issue #689, opened 2025-09-23
Pipewire source option stopped working with last update


Holbormon:
Hi, 

Before updating to the last release, the following config used to work to visualise the soundcard's microphone:

```
[input]
method = pipewire
source = alsa_input.pci-0000_00_1f.3.analog-stereo
```

But now, no sound is picked up anymore, as if `source` was incorrectly set. While debugging, I found out that the `source` entry above was technically set to `node.name` as printed by `wpctl inspect`, so I tried `object.path` and `object.serial` as stated by the README, but _doing so did not resolve the problem._

Under the current release, the only thing that results in proper sound pickup with this device is switching `method` to `pulse` and using `node.name`. Here is an excerpt of `wpctl inspect` on the device of concern:

```
node.name = "alsa_input.pci-0000_00_1f.3.analog-stereo"
node.nick = "ALC1220 Analog"
node.pause-on-idle = "false"
object.path = "alsa:acp:PCH:0:capture"
object.serial = "54"
```

OS: ArchLinux 6.16.8-arch2-1
`cava` package from official `extra` repositories

karlstav:
Any ideas @pallaswept? Didn't you make the changes here?

pallaswept:
> Any ideas [@pallaswept](https://github.com/pallaswept)? Didn't you make the changes here?

I did. I'll drop everything and get this working. I'll also subscribe to the issues here so I catch it if any others come up.

This is working here, though. OP's source seems to be correctly set, so I don't think that's the problem. A quick look at something like qpwgraph should confirm that, though.

@Holbormon does source `auto_input` function correctly? (it should capture sound from the default input device)

Did you try to set `active=1`? This is required for cava to capture audio from the device if no other applications are recording from it (otherwise the device will become idle and suspended until an active node is capturing from it, and then cava will show the audio which that application captures). Alternatively you could try just recording from the device and seeing if cava shows the audio then.

Is the device shown in `pw-top` with an 'R' in the left column? Is cava listed there as a follower (has a `+` before the name) of that device (listed below that device)? Does cava also have an 'R' in the left column? Is the number in the ERR column on both nodes, a 0?

We'll figure this out :)

Holbormon:
Thank you for the quick and detailed intervention ! 

First, I can confirm that `auto_input` does not change anything, and that `qpwgraph` **does** show that cava is listening to the right device.

Secondly, here's what happens when 1. recording from another device simultaneously and 2. adding `active = 1`:

[...]

PR #699, opened 2025-10-13
Terminate on error

pallaswept:
This will detect disconnection or error from the server, and terminate cava.

Prevents crashing on shutdown.

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -761,9 +761,10 @@ Keys:\n\
                     audio.terminate = 1;
                     exit(EXIT_FAILURE);
                 }
-                if (number_of_bars <= 1) {
+                if (number_of_bars < output_channels) {
                     cleanup();
-                    fprintf(stderr, "fixed number of bars must be at least 1\n");
+                    fprintf(stderr, "fixed number of bars must be at least 1 with mono output and "
+                                    "2 with stereo output\n");
                     exit(EXIT_FAILURE);
                 }
                 if (output_channels == 2 && number_of_bars % 2 != 0) {


pallaswept:
This is to solve the hang on shutdown reported [here](https://github.com/karlstav/cava/issues/689#issuecomment-3364402937)

pallaswept:
4 hours of running tests later I found an edge case. If it starts passive it will work as intended, but if it starts passive, becomes active, and then becomes passive, it will fail.

I have a model that passes that test but it was a bit more of a change so I avoided it. That's the second time I've made that mistake here. I'll need to make one more commit here to implement that, to catch that edge case.

pallaswept:
It turns out that it is actually outside of the pipewire code when it does this. The pipewire event handler sees the disconnection event, it sets terminate=1, it calls the function to quit the pipewire loop, and it returns from `input_pipewire` with the timers stopped and all... I'm not sure why it doesn't just quit with an error at that point like it does if it is active. It's still responsive, there is no output. but it requires the user to press Q to quit.

I'm not sure why it's doing that, I hope you might be able to take a look, and I'm going to have to brush up on my debugging in the meantime.

karlstav:
do you have the sleep_timer configured?

in that case it will not jump down to where it checks for the audio thread terminate status. 

https://github.com/karlstav/cava/blob/master/cava.c#L1087

maybe I should move it...

pallaswept:
That was it!

I was debugging it and I couldn't reproduce it in the debugger or with the debugger attached to it in a terminal, which was weird, I don't know if that makes any sense to you? Anyway I saw this, commented out that line, and the problem evaporated. Put it back, and it came right back, repeated three times. You did it! :tada: 

Funny thing about that - the downstream project that made me notice this bug, kurve, has a GUI and it restricts us to a sleep_timer value > 0, so I was guaranteed to trip this one all the way :D I edited that on my PC just now, and omitted the sleep_timer, and saw a few problems evaporate.

I hope you wouldn't mind taking a look at that, I wouldn't feel confident stepping out there into the rendering code, I've been stepping through it for a few hours in gdb and I know it's above my paygrade :D I might update the above code after, so I can be sure it will behave correctly.

Thanks again Karl!

karlstav:
I have a commit ready to fix it. Will push it later.

PR #701, opened 2025-10-25
pipewire use pointer to struct instead of struct

LukashonakV:
Dear @karlstav , 
Because of the possibility to use cava as a shared library where C++ application tries to use it it's small memory leak when pipewire input method is used. The reason - internal local struct which is managed in the input_pipewire method.
This PR just replaces direct struct variable with the pointer which does nothing to cava itself but prevents memory leaking when external applications using cava as a library. 
May you please consider this a small improvement of the library. 
Thanks

### DIFF
--- a/input/pipewire.c
+++ b/input/pipewire.c
@@ -136,34 +136,32 @@ unsigned int next_power_of_2(unsigned int n) {
 }
 
 void *input_pipewire(void *audiodata) {
-    struct pw_data data = {
-        0,
-    };
+    struct pw_data *data = (struct pw_data *)malloc(sizeof(struct pw_data));
 
-    data.cava_audio = (struct audio_data *)audiodata;
+    data->cava_audio = (struct audio_data *)audiodata;
     const struct spa_pod *params[1];
-    uint8_t buffer[data.cava_audio->input_buffer_size];
+    uint8_t buffer[data->cava_audio->input_buffer_size];
     struct pw_properties *props;
     struct spa_pod_builder b = SPA_POD_BUILDER_INIT(buffer, sizeof(buffer));
     uint32_t nom;
-    nom = next_power_of_2((512 * data.cava_audio->rate / 48000));
+    nom = next_power_of_2((512 * data->cava_audio->rate / 48000));
     pw_init(0, 0);
 
-    data.loop = pw_main_loop_new(NULL);
-    if (data.loop == NULL) {
-        data.cava_audio->terminate = 1;
-        sprintf(data.cava_audio->error_message,
+    data->loop = pw_main_loop_new(NULL);
+    if (data->loop == NULL) {
+        data->cava_audio->terminate = 1;
+        sprintf(data->cava_audio->error_message,
                 __FILE__ ": Could not create main loop. Is your system running pipewire? Maybe try "
                          "pulse input method instead.");
         return 0;
     }
 
-    data.timer = pw_loop_add_timer(pw_main_loop_get_loop(data.loop), on_timeout, &data);
+    data->timer = pw_loop_add_timer(pw_main_loop_get_loop(data->loop), on_timeout, data);
 
     props = pw_properties_new(PW_KEY_MEDIA_TYPE, "Audio", PW_KEY_MEDIA_CATEGORY, "Capture",
                               PW_KEY_MEDIA_ROLE, "Music", NULL);
 
-    char *source = data.cava_audio->source;
+    char *source = data->cava_audio->source;
     size_t sourcelength = strlen(source);
 
     if (8 <= sourcelength && !strcmp(source + sourcelength - 8, ".monitor")) {
@@ -175,19 +173,19 @@ void *input_pipewire(void *audiodata) {
     } else if (strcmp(source, "auto_input") != 0) {
         pw_properties_set(props, PW_KEY_TARGET_OBJECT, source);
     }
-    pw_properties_setf(props, PW_KEY_NODE_LATENCY, "%u/%u", nom, data.cava_audio->rate);
+    pw_properties_setf(props, PW_KEY_NODE_LATENCY, "%u/%u", nom, data->cava_audio->rate);
 
-    if (data.cava_audio->active)
+    if (data->cava_audio->active)
         pw_properties_set(props, PW_KEY_NODE_ALWAYS_PROCESS, "true");
     else
         pw_properties_set(props, PW_KEY_NODE_PASSIVE, "true");
 
-    if (data.cava_audio->virtual_node)
+    if (data->cava_audio->virtual_node)
         pw_properties_set(props, PW_KEY_NODE_VIRTUAL, "true");
 
     enum spa_audio_format audio_format = SPA_AUDIO_FORMAT_S16;
 
-    switch (data.cava_audio->format) {
+    switch (data->cava_audio->format) {
     case 8:
         audio_format = SPA_AUDIO_FORMAT_S8;
         break;
@@ -202,22 +200,22 @@ void *input_pipewire(void *audiodata) {
         break;
     };
 
-    if (data.cava_audio->remix) {
+    if (data->cava_audio->remix) {
         pw_properties_set(props, PW_KEY_STREAM_DONT_REMIX, "false");
         pw_properties_set(props, "channelmix.upmix", "true");
 
-        if (data.cava_audio->channels < 2) {
+        if (data->cava_audio->channels < 2) {
             // N to 1 with all channels shown
             params[0] = spa_format_audio_raw_build(
                 &b, SPA_PARAM_EnumFormat,
-                &SPA_AUDIO_INFO_RAW_INIT(.format = audio_format, .rate = data.cava_audio->rate,
-                                         .channels = data.cava_audio->channels, ));
+                &SPA_AUDIO_INFO_RAW_INIT(.format = audio_format, .rate = data->cava_audio->rate,
+                                         .channels = data->cava_audio->channels, ));
         } else {
             // N to 2 with all channels shown
             params[0] = spa_format_audio_raw_build(
                 &b, SPA_PARAM_EnumFormat,
-                &SPA_AUDIO_INFO_RAW_INIT(.format = audio_format, .rate = data.cava_audio->rate,
-                                         .channels = data.cava_audio->channels,
+                &SPA_AUDIO_INFO_RAW_INIT(.format = audio_format, .rate = data->cava_audio->rate,
+                                         .channels = data->cava_audio->channels,
                                          .position = {SPA_AUDIO_CHANNEL_FL,
                                                       SPA_AUDIO_CHANNEL_FR}, ));
         }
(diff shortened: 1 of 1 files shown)


input/pipewire.c
```diff
@@ -202,53 +200,56 @@ void *input_pipewire(void *audiodata) {
...
 
     if (status < 0) {
-        data.cava_audio->terminate = 1;
-        sprintf(data.cava_audio->error_message,
+        data->cava_audio->terminate = 1;
+        sprintf(data->cava_audio->error_message,
                 __FILE__ ": Could not connect stream. Is your system running pipewire? Maybe try "
                          "pulse input method instead.");
-        return 0;
+        goto error_free;
```
karlstav: this could be
```
else {
    pw_main_loop_run(data->loop);
    pw_stream_destroy(data->stream);
}
pw_loop_destroy_source(pw_main_loop_get_loop(data->loop), data->timer);
...
```
could it not?

karlstav:
how does making it a pointer an allocating freeing it prevent memory leak? Maybe this is just me not understanding how cava is implemented as a library.

input/pipewire.c
LukashonakV: Yes, good point.  Let me recheck once I reach PC

LukashonakV:
Hi @karlstav , please take it on pause. Actually I don't have an explanation right now. And now I end up with the ideas.
How now it's working.
It's Waybar is written in C++. It uses libcava (which is very similar to the origin cava project in order to do it simple merge changes from  the upstream)
So... libcava is extended with the
https://github.com/LukashonakV/cava/blob/ddeecc8a2c9f786a03ff3ce2a29918573ef9c42e/include/cava/input/common.h#L51-L52
Idea to provide method to the external applications input method is based on the provided config.. Currently method represents the  pointer to particular input method(pulse, pipewire, etc.).
Then Waybar declares libcava as a extern C
https://github.com/LukashonakV/Waybar/blob/d72f38b5c7889224de724fda1fd82b760c2401df/include/modules/cava/cava_backend.hpp#L8-L21
Declares pointer to the input_method
https://github.com/LukashonakV/Waybar/blob/d72f38b5c7889224de724fda1fd82b760c2401df/include/modules/cava/cava_backend.hpp#L46
Then during initialization step gets the method itself
https://github.com/LukashonakV/Waybar/blob/d72f38b5c7889224de724fda1fd82b760c2401df/src/modules/cava/cava_backend.cpp#L229
And finally starts reading in separate thread
https://github.com/LukashonakV/Waybar/blob/d72f38b5c7889224de724fda1fd82b760c2401df/src/modules/cava/cava_backend.cpp#L15-L27

----

So far all looks quite correct. 
I also checked cava input_pipewire many times... and don't see any issues here. More over when I compile and run libcava separately valgrind doesn't detect any memory leaks... but when waybar is compiled and run valgrind says
```
==25169== 32 bytes in 1 blocks are definitely lost in loss record 6,656 of 13,858
==25169==    at 0x4B96EF3: calloc (vg_replace_malloc.c:1675)
==25169==    by 0x631BB59: pw_main_loop_new (in /usr/lib64/libpipewire-0.3.so.0.1408.0)
==25169==    by 0x4BC3D86: input_pipewire (pipewire.c:152)
==25169==    by 0x42B5681: operator() (cava_backend.cpp:26)
==25169==    by 0x42B5681: __invoke_impl<void, waybar::modules::cava::CavaBackend::CavaBackend(const Json::Value&)::<lambda()>&> (invoke.h:61)
==25169==    by 0x42B5681: __invoke_r<void, waybar::modules::cava::CavaBackend::CavaBackend(const Json::Value&)::<lambda()>&> (invoke.h:111)
==25169==    by 0x42B5681: std::_Function_handler<void (), waybar::modules::cava::CavaBackend::CavaBackend(Json::Value const&)::{lambda()#1}>::_M_invoke(std::_Any_data const&) (std_function.h:290)
==25169==    by 0x40C59E0: operator() (std_function.h:591)
==25169==    by 0x40C59E0: operator() (sleeper_thread.hpp:48)
==25169==    by 0x40C59E0: __invoke_impl<void, waybar::util::SleeperThread::operator=(std::function<void()>)::<lambda()> > (invoke.h:61)
==25169==    by 0x40C59E0: __invoke<waybar::util::SleeperThread::operator=(std::function<void()>)::<lambda()> > (invoke.h:96)
==25169==    by 0x40C59E0: _M_invoke<0> (std_thread.h:301)
==25169==    by 0x40C59E0: operator() (std_thread.h:308)
==25169==    by 0x40C59E0: std::thread::_State_impl<std::thread::_Invoker<std::tuple<waybar::util::SleeperThread::operator=(std::function<void ()>)::{lambda()#1}> > >::_M_run() (std_thread.h:253)
==25169==    by 0x664D5C3: ??? (in /usr/lib/gcc/x86_64-pc-linux-gnu/14/libstdc++.so.6.0.33)
==25169==    by 0x6986E63: ??? (in /lib64/libc.so.6)
==25169==    by 0x6A076B3: clone (in /lib64/libc.so.6)
```
I even checked pipewire source code and it also look Ok. 
https://github.com/LukashonakV/cava/blob/ddeecc8a2c9f786a03ff3ce2a29918573ef9c42e/src/input/pipewire.c#L251C5-L251C37 - loop is really should be free
For now I tried C++ wrappers but not luck. But I noticed that pulse for example doesn't provide a cause for memory leak. When I checked it - internally it uses pointer to struct... whereas pipewire uses struct istelf. So when I replaces struct to pointer to struct this issue is gone. But I can't explain why. AI also says the event existing chain waybar and libcava should not produce memory leaks. My suspicious - it's some peculiarity here that I don't get related to the managing of the functionalities between C++ and C

And finally report from AI

----
Based on your description, the key difference is that when you declare struct pw_data data as a local variable, it is allocated on the stack, and Valgrind reports a leak because the code calls functions that allocate memory internally (e.g., pw_main_loop_new, pw_stream_new_simple, etc.), but these are destroyed only at the end of the function — so Valgrind sees that the internal allocations are not freed explicitly.

When you replace the stack allocation with a pointer to a dynamically allocated struct pw_data, the leak report disappears, likely because:

    The memory for the pw_data struct itself is freed explicitly or at least the program's exit cleans it up.
    Also, the internal resources (main loop, stream, source) are cleaned up explicitly at the end (pw_stream_destroy, pw_loop_destroy_source, pw_main_loop_destroy), so Valgrind doesn't report leaks.

How to properly fix the leak when using stack allocation:

    Ensure all internal resources are properly freed before the function returns.
        You are already doing this in your error_fallback label:

                  

        if (data.stream)
            pw_stream_destroy(data.stream);
        if (data.timer)
            pw_loop_destroy_source(pw_main_loop_get_loop(data.loop), data.timer);
        if (data.loop)
            pw_main_loop_destroy(data.loop);
        pw_properties_free(props);
        pw_deinit();

              

    Make sure this cleanup code runs regardless of errors.
    Use a goto cleanup pattern or a cleanup function to guarantee cleanup on all paths.

Important details:

    The internal resources (loop, stream, timer) are dynamically allocated by PipeWire functions.
    You must explicitly free/destroy these resources.
    The struct itself (pw_data) is on the stack, so no need to free it unless you allocated it dynamically.
    Valgrind reports leaks if:
        Some resources are allocated but not freed.
        Or, the cleanup code is bypassed on some error paths.

Summary

    When using stack allocation: Make sure all internal resources (loops, streams, timers) are destroyed explicitly before the function returns.
    When switching to dynamic allocation (pointer): You need to delete or free the pw_data pointer after cleanup, but the leak report may decrease because the memory management is clearer.

In your case:

    Keep your current pattern.
    Add cleanup code for all allocated resources.
    Ensure the cleanup code runs on all paths (success, error).

If you'd like, I can help you draft a robust cleanup pattern or modify your code to ensure no leaks.

karlstav:
Hmm long read, I don't really have time to go into the details.

I can merge it if you like. It seem to work just as well as before.

BTW I think the proper way to move forward in general would be to expand the cavacore library instead of modifying cava into a "libcava". You can utilise the CMake stuff already in the code, add the config file loader stuff and input/output modules as needed. I would be willing to move more code out from main and into other files that can be added to cavacore library, if needed.

LukashonakV:
> BTW I think the proper way to move forward in general would be to expand the cavacore library instead of modifying cava into a "libcava".

Agree with you. So. Let's open an ISSUE then where I can put a plan ,, and according the plan will start providing pull requests ... where finally extended cavacore could incorporate all features are implemented in libcava ? 
If you agree with that.. I'll create such one

karlstav:
> where finally extended cavacore could incorporate all features are implemented in libcava ?

yes exactly! just go ahead and create the issue.

should I still merge this?

pallaswept:
I have not had an opportunity to test this yet, but just looking at the code I should point out that there is a significant difference in functionality between these two variants - the original initialises the struct, this commit does not, it only allocates it.

Another, perhaps the reason for the unexplained remedy of the unexplained leak, is that in addition to changing from struct to pointer,`pw_loop_destroy_source(pw_main_loop_get_loop(data->loop), data->timer);` is added.

This is related to 6e4fda158de66bbf5cd33241605ed31e051a1232 That timer was newly added in my recent changes (to keep the loop responsive when the node is idle/suspended), and the original doesn't explicitly destroy it when it's done (this also caused problems when the server crashed - it would be cleaned up when the loop is destroyed, but that might not happen from pw's side), but this commit does that cleanup explicitly now (better, thanks OP). That, would mean that the AI was half-right, which is usually how they work :smile: 

Has the original code, with that added cleanup of the timer, been tested for the memory leak in the cpp application?

I think it's worthwhile to spend the time to understand the nature of the bug (what is even leaking?) and the fix (why isn't it leaking now?) what are the effects of the entirely new code (destroy the timer, should be better I think?) and if the change to a pointer is taken, to ensure that the struct is fully initialised otherwise.

And don't forget the goto :laughing: 

Just some quick observations and my opinion, but take it with a grain of salt, I'm not bothered so long as it keeps running so nicely :)

LukashonakV:
Thank you @pallaswept for the comments. Let's go one by one.

- Actually I'm not pushing nor @karlstav  neither @pallaswept  about this. The reason, I don't really like to put, take commits, and take it as is if I don't have clear explanation this or that thing is about. 
- I hope you guys clearly understand that code from commit is not generated or suggested by AI  :) AI I was used first time and the only like a last resort to have a clear explanation about this memory leak
- Unfortunately thiis is ```pw_loop_destroy_source(pw_main_loop_get_loop(data->loop), data->timer);```not a remedy
- I've already tested provided PR with C++ and memory leak is disappeared. 
- What is I see now, looks like memory leak is happening when the instance goes to destruction phase. According valgrind the memory is allocated here
```==9417== 32 bytes in 1 blocks are definitely lost in loss record 6,651 of 13,886
==9417==    at 0x4B96EF3: calloc (vg_replace_malloc.c:1675)
==9417==    by 0x631BB59: pw_main_loop_new (in /usr/lib64/libpipewire-0.3.so.0.1408.0)
==9417==    by 0x4BC3D86: input_pipewire (pipewire.c:152)
```
is lost.
- From the previous point I found interesting dependency... according the input_pipewire code base looks like all memory free should be done and no memory leaks should happen. But when I comment for test purposes string ```pw_main_loop_run(data.loop);``` issue with memory leak is disappear. Which hints that inside pw_main_loop_run memory allocation is happening as well. Even it the end ```pw_main_loop_destroy(data,loop);``` - this should free the memory .. valgrind detects memory leaking
- Taking in account is saying above and checking AI report... which also hints that looks like the different due to struct itself is keeping in stack... where pointers are in heap (which for a moment I totally forgot about) I'm trying to check how exactly destruction (with related memory frees) is happening in C++ and C.

pallaswept:
> * I hope you guys clearly understand that code from commit is not generated or suggested by AI  :)

Of course! :) I have the impression that you just happened to ask the AI, after you were finished. It can't hurt to ask, right :) I think they can be very useful tools.

> Unfortunately thiis is pw_loop_destroy_source(pw_main_loop_get_loop(data->loop), data->timer);not a remedy

I don't understand why it is added if it was not intended to fix this problem and is not used in the documentation. What was the motivation and the source of this? It seems like it came from nowhere for no reason and does nothing :D I feel like I have missed something.

> * According valgrind the memory is allocated here 
> ```
> ==9417==    at 0x4B96EF3: calloc (vg_replace_malloc.c:1675)
> ==9417==    by 0x631BB59: pw_main_loop_new (in /usr/lib64/libpipewire-0.3.so.0.1408.0)
> ==9417==    by 0x4BC3D86: input_pipewire (pipewire.c:152)
> ```

Perhaps you have found a pipewire bug exposed by cava?

> * From the previous point I found interesting dependency... according the input_pipewire code base looks like all memory free should be done and no memory leaks should happen. But when I comment for test purposes string `pw_main_loop_run(data.loop);` issue with memory leak is disappear. Which hints that inside pw_main_loop_run memory allocation is happening as well. Even it the end `pw_main_loop_destroy(data,loop);` - this should free the memory .. valgrind detects memory leaking

This seems like a pipewire bug again?

The developers there are very helpful and will take a report like this seriously if it can be reproduced. For example look to the pipewire reference: https://docs.pipewire.org/devel/audio-capture_8c-example.html
They use the struct, not the pointer to the struct, there. It seems that they intend for it to work.

You've discovered something very interesting. I wonder what it is :D

Issue #703, opened 2025-10-28
Adjustement with libcava


LukashonakV:
Idea for this issue is to have the plan about what and how should be extended in existing cava project to have an ability to be compiled as a shared library with providing of the necessary API which can be used by 3-rd party projects. Shod be kept till plan is implemented.
@karlstav  FYI (if any doubts, disagreements, let's have a talk)

- [x] To organize in common way project files. Header in include directory, source files in src
-- PR #705
- [x] To provide meson configuration. Existing cmake and meson should be updated simultaneously
- [x] To move common blocks in the source files into separate useful methods
-- PR #706
- [x] To adjust headers with libcava
-- PR #706 
- [ ] To close libcava project. Move waybar dependency to the original cava project

karlstav:
ok, so this was not exactly what I had in mind. Reorganizing everything feels great, but my philosophy is more like "do what little is needed to achieve the goal at hand". Meson is probably awesome, but we are already stuck with cmake due to the android stuff, I will not add another build system as discussed in other issues.

This was my plan:

1) add config.c/h and required input modules to cavacore in cmake (make it optional)
2) move required functionality from cava.c to cavacore.c (or new files and add them to cavacore via cmake).
3) make modifications to existing files if needed, but as little as absolutely possible.

LukashonakV:
@karlstav  (1,2) points are covered by the 3,4 from the plan above. But can we discuss a little, to clarify.
1. Just reorganizing the file structure actually does not bring any harm. Even CI shows it. In the same time at leas a couple of points could be considered as an advantages like: more clear structure where headers are placed in include folder where in the future for example could be separate once which will contains only Shared header which will provide necessary API in library style. And this exact folder can be considered as a source to deploy this set of API into the target system. Even looking at this, the view much more structures and simplified 

<img width="1382" height="862" alt="Image" src="https://github.com/user-attachments/assets/7f870163-4165-4e73-96c5-c17e91b2cd57" />
2. meson .... why is this so principal ? It will worth almost nothing to have just one more build system because of meson in libcava already written, working well and for that even separate CI is configured already. The only thing is needed to change each time cava version is promoted, to change version value right in the beginning of the meson configuration
Example of the existing meson configuration: https://github.com/LukashonakV/cava/blob/master/meson.build

<img width="1061" height="697" alt="Image" src="https://github.com/user-attachments/assets/c284b4ed-754d-4510-ae4b-abaa4717301a" />

[...]

PR #706, opened 2025-11-05
Adjustement with libcava

LukashonakV:
Dear @karlstav ,
This PR covers two points from the #703

1. To move common blocks in the source files into separate useful methods
2. To adjust headers with libcava

----

So it's huge work routine is done
1. Original files were not restructured
2. Common blocks are moved under separate functionalities
3. headers we reconfigured in order to segregate those part of headers which should provide shared API for external applications. The other part should be internal
4. It's separate include folder which contains headers and only open API. This is also considered to be installed into the system
5. configure.ac and Makefile.am are rewritten in order to support application and libcava building, linking installation
6. It's tested. application is working well on my machine, libcava is also is installed into the system. External application waybar is available to access and use libcava API
7. Actually Makefile.am with configure.ac was a headache to me. Meson much more simple. Even library itself in meson take less size.

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -163,7 +163,6 @@ Keys:\n\
     // general: main loop
     while (1) {
 
-        debug("loading config\n");
         // config: load
         struct error_s error;
         error.length = 0;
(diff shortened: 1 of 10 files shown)


LukashonakV:
>Covers Alexays/Waybar#4533

Issue #711, opened 2025-12-10
[Windows] Add the ability to choose the audio device


Deide:
I have multiple audio devices with one of them being dedicated to music. I want to run cava in my top bar, but I want it responding to my music, not my system or communications audio.

The windows build picks the default device and it isn't configurable. It would be nice to be able to set the device in the config.

pallaswept:
> and it isn't configurable.

Windows provides an interface to change audio device per-application. Give it a shot. GL!

Deide:
> Windows provides an interface to change audio device per-application. Give it a shot. GL!

Don't know why I didn't think of this! It works!
But then I restart cava, and in the Windows Settings audio router, it still reports that it's set to the correct device, but cava hears nothing until I toggle it back and forth between another device and the one playing music.

I've had some other (old) programs act this way too.
I'm on Windows 10, so maybe it's no longer an issue in Windows 11.

pallaswept:
Well, darn. I'm afraid that Windows' audio engine is a large part of the reason I'm a linux user these days,  so I'm a bit out of the loop on the current state of Windows' sound stuff. I was kinda hoping I had a quick fix for you but I better leave it to smart people like Karlstav :)

karlstav:
Hehe, I am not that hard on the windows either. We could try @rohanb1216 who has done the most work here recently.

rohanb1216:
What should be the behavior if the specified device is not found?

karlstav:
Exit with error I think would be best

karlstav:
closing this as it was implemented.

Nice work @rohanb1216!

PR #712, opened 2025-12-13
Device selection from config win

rohanb1216:
Implements #711


karlstav:
looks good and sorry about the delay.

I'm just going to trust that this works. No time to boot into windows.

rohanb1216:
Just realized I bunged up the check to see if a device was specified or not (looked for a null string but it's populated with 'auto' instead)

PR #728, opened 2026-01-15
input/pipewire: satisfy clang-format case formatting

oxyzenQ:
This change adjusts the PipeWire stream state switch-case formatting to satisfy clang-format 9.
 
No functional change intended; this is purely formatting to keep CI/lint happy.
 
Testing:
- Build: make -j
- Basic runtime: run cava with pipewire input enabled


karlstav:
hmm, I guess it complains because a new variable is declared inside the case switch, making the scope of it sort of ambiguous.

Issue #727, opened 2026-01-15
SDL_GLSL: add sdl_glsl_gain and improve long-run stability + reduce overhead


oxyzenQ:
### Describe the bug

When using `output_method = sdl_glsl`, the `shader_time` uniform is derived from a frame counter. Over long runtimes this can lead to reduced float precision and visible instability in time-based GLSL shaders (e.g. subtle jitter/stutter in animations).

Additionally, the SDL_GLSL backend currently sets up and uses an FBO + a second render pass unconditionally, even when the selected shader does not use the `inputTexture` uniform. This adds unnecessary GPU work and can increase GPU cost/latency for shaders that only need the current bar data.

### To Reproduce

1. Build cava from source (master) with SDL and OpenGL enabled.
2. Use a config like:

```ini
[general]
output_method = sdl_glsl

[sdl]
shader = orion_circle_rotate   # or any shader that uses shader_time
# continuous_rendering = 1     # if you want guaranteed animation updates
```

3. Run:

```sh
cava -p /path/to/config
```

4. Let it run for a long time (hours) and observe that animation smoothness can degrade (jitter/stutter) due to `shader_time` precision issues.

5. For the FBO overhead case:
   - Set `shader = orion_circle` (a shader that does **not** sample `inputTexture`)
   - Run cava and observe that SDL_GLSL still does the extra FBO/second-pass path despite `inputTexture` being unused.

### Expected behavior

- `shader_time` should remain stable over long runtimes (monotonic, no precision drift affecting animation smoothness).
- SDL_GLSL should only allocate/use the FBO + second pass when the shader actually uses `inputTexture`.

### Screenshots

N/A (timing/overhead issue).

### Desktop (please complete the following information)

- OS: `Archlinux`
- Version: `Linux nightpc 6.18.5-1-workstation-threadripper-eevdf-lto`
- GPU: `AMD Radeon Pro W7900 (Mesa 25.3.3, RADV)`
- Cava version/commit: `e8bf686 (sdl_glsl: use SDL perf counter for shader_time)`

### Terminal emulator

- Name: `Alacritty`
- Version: `alacritty 0.16.1 (42f49eeb)`

### Additional context

This affects long-endurance use and stable frame pacing for time-based shaders. The unnecessary FBO pass also impacts GPU cost even for shaders that only use `bars` and do not sample `inputTexture`.

PR #730, opened 2026-01-15
SDL_GLSL: add gain option, reduce overhead, and harden long-run timing

oxyzenQ:
Fixes https://github.com/karlstav/cava/issues/727

SDL_GLSL improvements:
- Add output:sdl_glsl_gain (percent) to scale bar amplitudes before they are sent to shaders.
- Improve long-run stability of shader_time using SDL performance counter + wrapping to avoid float precision drift.
- Skip FBO setup and the extra render pass when the shader does not use inputTexture.
- Keep clang-format-lint passing for output/sdl_glsl.c changes.
- Clarify example_files/config comments for continuous_rendering, sdl_glsl_gain, and disable_blanking.

Testing:
- Build: make -j
- Runtime: run cava with output method sdl_glsl and confirm:
  - sdl_glsl_gain affects intensity as expected
  - shaders that don’t use inputTexture still work
  - long sessions don’t degrade shader_time behavior

### DIFF
--- a/example_files/config
+++ b/example_files/config
@@ -226,12 +226,15 @@
 ; vertex_shader = pass_through.vert
 ; fragment_shader = bar_spectrum.frag
 
-; for glsl output mode, keep rendering even if no audio
+# For glsl output mode, keep rendering even if no audio. 1 = on, 0 = off. Default: 0
+# Enable for animated shaders so time-based effects keep updating.
 ; continuous_rendering = 0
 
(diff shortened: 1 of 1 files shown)


karlstav:
nice to see someone take interest in the sdl_glsl output, I wrote it wanting to learn more about shaders, but wrapping my head around the math of it is taking some time.

PR #731, opened 2026-01-17
fix(win-installer): update post-installation instructions and add PATH variable modification

amnweb:
- Added instructions for the user regarding the installation directory being added to the PATH environment variable.
- Included a note about the need to restart the terminal or log out for changes to take effect.
- Updated the installer script to modify the user's PATH environment variable to include the installation directory.

### DIFF
--- a/cava_win/postsetupinfo.txt
+++ b/cava_win/postsetupinfo.txt
@@ -1,5 +1,8 @@
 config file for cava is created in "%USERPROFILE%\.config\cava" upon first launch
 
+The installation directory has been added to your PATH environment variable.
+You may need to restart your terminal or log out and log back in for the 'cava' command to work.
+
 you need to install Visual C++ redistributable if you do not already have it:
 
 https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist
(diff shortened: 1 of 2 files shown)


amnweb:
There is a problem because once the path is added, it won’t be removed on uninstall. And this Inno Setup doesn’t have an option for it, or I can’t find it in the docs.

Are you interested in switching this Inno Setup to Wix and MSI build and also integrating it into the workflow so the build will run fine?

karlstav:
maybe, but I love Inno setup, it looks like it was written by one guy 30 years age and it still works like a charm!

the Wix thing looks like it costs money? does it? Having it integrated into the worflow would be cool. Could it upload the installer as an artifact? Then I wouldn't have to boot into stupid windows

amnweb:
Yeah, I already built the MSI locally and it works fine. I will submit a PR in the next few days. Yes, the MSI installer as an artifact file will be uploaded, so you do not need to use Windows.

amnweb:
I have closed this PR. I’m working on the MSI installer right now.

mxwrld:
Ok thank you, I'll keep checking on updates haha.
By the way, is it normal that with every my pc restart, the YASB update keep popping up.
Unless you still making some changes, but if it's not then what should i do, i already installed it 3 times with every boot ( through option "About" and "Check for updates" .

PR #733, opened 2026-01-17
sdl: avoid recreating window on config reload

oxyzenQ:
This change avoids recreating the SDL/SDL_GLSL window on live config reload when only
"safe" parameters change (e.g. bar_width / bar_spacing), preventing the window from
jumping to the active workspace.

- Add a soft-reload path for OUTPUT_SDL and OUTPUT_SDL_GLSL:
  - Re-init surfaces and update runtime parameters without destroying the window
  - If bar_width/bar_spacing changes, trigger layout recalculation (resizeTerminal)
- If load_config() temporarily fails during atomic save, keep running with old config
  and retry next loop instead of doing a full reload.

Tested:
- Built successfully (make)
- Verified live reload no longer causes window jump on small config edits

Signed-off-by: rezky_nightky <with.rezky@gmail.com>

Author: rezky_nightky <with dot rezky at gmail dot com>
Repository: cava
Branch: pr/window-jump-reload
Signing: GPG (4B65AAC2)
HashAlgo: BLAKE3

[ Block Metadata ]
BlockHash: 23e1afcf36367ea00675f2ce407ef03d5b1b197ba1c2c655a75e5b774d05011f PrevHash: 8192c30da5cc58cf3450d8cd994133ce4c62eca0ab944e8d420e076adbad8a97 PatchHash: c74ee507d5831493bb6388108d30acd3342265878fc63c00023a525ad936e61a

FilesChanged: 1
Lines: +166 / -0

Timestamp: 2026-01-17T18:27:30Z
Signature1: 73965e24aa8b42dc44375f739b85042e7d5dc11a43918f1701aab266b7cfbee0
Signature2: f742cd8e566773b3afcb43a396d891d8dbe0a8a1067c3d2df2a6070ed90ac0e5

### DIFF
--- a/FORMATTING.md
+++ b/FORMATTING.md
@@ -31,11 +31,11 @@ If you see a warning like:
 
 `Warning: scripts prefer clang-format 9 ...`
 
-you can avoid CI/local differences by installing `clang-format-9` (preferred) or installing `docker`.
+you can avoid version differences by installing `clang-format-9` (preferred) or installing `docker`.
 
 ## Docker notes (Linux)
 
-If you want to use the Docker-based clang-format 9 (CI-matching) path:
+If you want to use the Docker-based clang-format 9 path:
 
 1. Start the Docker daemon:
 


karlstav:
hmm I'm sorry, but you will have to find a way to do this  that does not dump 167 lines of code into the main loop...

the code is already not very tidy (actually doing some clean up at the moment) but this does not help

oxyzenQ:
@karlstav Cleaned to a single commit on top of upstream/master.

Behavior is preserved (avoid recreating SDL/SDL_GLSL window on config reload when “safe”),
but refactored so the main loop stays small/reviewable:
- moved soft-reload decision + apply logic into helper functions
- centralized frame timing recalculation into helper

Build sanity check: make (OK).

karlstav:
still seems like a lot of code. Maybe we could just not destroy/recreate the sdl window on config reload? It would brake some of the reload functionalities when in sdl output, but I am not sure if it matters. Would be a lot more clean code.

oxyzenQ:
@karlstav I simplified the SDL/SDL_GLSL live reload logic to avoid window recreation entirely. On reload we always load_config() into a temporary params struct and then apply only a small safe subset of runtime parameters (sens/framerate/bar_width/bar_spacing + colors/gradient + SDL_GLSL continuous_rendering/gain). Other config changes still require restart. For SDL_GLSL, shader path changes now trigger an in-place shader program rebuild (no window re-create), while keeping a stable attribute location to avoid VAO issues.

karlstav:
I'm still a bit skeptical. The reason being that this is a hack. It breaks the regular logic of the application and makes the code harder to manage down the road. There are already tons of hacks, but I just don't want to add more.

I do see the need for this, maybe we as a compromise can instead put the parameters in question into the "reload_colors" function?  So that in addition to reloading colors we also reload:

-   p->sens 
-   p->framerate
-   p->continuous_rendering
-   p->bar_width
-   p->bar_spacing
-   p->sdl_glsl_gain
-   p->vertex_shader
-   p->fragment_shader

I don't think we have to rename the reload colors functionality, just add a note that it also reloads other "superficial drawing parameters", or something similar.

Would that work? reloading colors would trigger "init_sdl_surface", but not "init_sdl_window"

oxyzenQ:
@karlstav Hi Karl,

I’ve pushed an updated version to the PR branch (force-pushed `pr/window-jump-reload` to commit aacd48d).

As suggested, SDL/SDL_GLSL reload now reuses the existing `reload_colors` path to also reload the “superficial drawing params” in-place (sens/framerate/continuous_rendering/bar_width/bar_spacing/sdl_glsl_gain), and SDL_GLSL hot-reloads shaders when the vertex/fragment shader paths change — without recreating the SDL window (no jump).

Local checks passed:
- make format-check
- make -j4 CFLAGS='-Werror'

Thanks!

karlstav:
looks alright, not sure about the formatting thing though. Is it really the job of the build system? Most IDEs do this by themselves now. If not, people can always dig up some tool for it themselves.

oxyzenQ:
@karlstav Hi Karl,

Makes sense — agreed that formatting shouldn’t be part of the build system/CI.

I’ve updated the PR accordingly:
- Removed the `make format` / `make format-check` targets from [Makefile.am](cci:7://file:///home/rezky/Desktop/cava/Makefile.am:0:0-0:0)
- Removed the clang-format lint job from [.github/workflows/build-and-test.yml](cci:7://file:///home/rezky/Desktop/cava/.github/workflows/build-and-test.yml:0:0-0:0)
- Kept the helper scripts for local use ([scripts/clang-format.sh](cci:7://file:///home/rezky/Desktop/cava/scripts/clang-format.sh:0:0-0:0) + [scripts/clang-format-lint.sh](cci:7://file:///home/rezky/Desktop/cava/scripts/clang-format-lint.sh:0:0-0:0)) and updated [FORMATTING.md](cci:7://file:///home/rezky/Desktop/cava/FORMATTING.md:0:0-0:0) to reference the scripts directly (no more make targets / CI wording)

Core PR behavior is unchanged: SDL/SDL_GLSL live reload continues to reuse the existing reload_colors path (superficial params only) and avoids recreating the window (no “jump”), with GLSL shader hot-reload still working.

Sanity:
- [./configure](cci:7://file:///home/rezky/Desktop/cava/configure:0:0-0:0) (no `-Werror`)
- `make -j4 CFLAGS='-Werror'` (OK)

Force-pushed to `pr/window-jump-reload` (HEAD: f55f443).

Thanks!

karlstav:
I really don't want any of the formatting stuff at all.

this file is the only thing the repo needs:

https://github.com/karlstav/cava/blob/master/.clang-format

Everything else is up to the IDE or what ever tools people chose to use. 

I could maybe add a note in the CONTRIBUTING.md that there is a .clang-format and that PRs will be not be merged if they don't pass the linter

PR #736, opened 2026-01-22
chore(config): improve config readability and consistency

oxyzenQ:
- Document default values directly in config comments
- Improve example formatting and clarity
- Align example defaults with actual config defaults (e.g. higher_cutoff_freq = 8000 on Linux)

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -89,26 +89,70 @@
 #ifdef _WIN32
 char *optarg = NULL;
 int optind = 1;
+int optopt = 0;
 
 static int getopt(int argc, char *const argv[], const char *optstring) {
     if ((optind >= argc) || (argv[optind][0] != '-') || (argv[optind][0] == 0)) {
         return -1;
     }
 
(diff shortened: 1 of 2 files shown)


karlstav:
actually the default value is always the commented out value in the config (except where stated otherwise), so this commit is a bit redundant.

there used to be a note about this though, I will re add it.

oxyzenQ:
The motivation was mainly for new users — the implicit rule isn’t obvious at first glance, especially when values are boolean-like (0/1, on/off).

Re-adding the note sounds like a good middle ground 👍
Thanks for taking a look.

oxyzenQ:
@karlstav the https://github.com/karlstav/cava/commit/5bbc2a664f3ea8552eb08c21f12735cbe64f28cc revert but with little improve.

oxyzenQ:
@karlstav Closing this since it no longer contains any changes.
Thanks!

PR #741, opened 2026-01-25
cli: make long options strict (no abbrev prefixes)

oxyzenQ:
Previously, long option parsing on non-Windows followed the default getopt_long behavior,
which allows unique prefixes for long options. As a result, some typos were accepted
silently, leading to inconsistent behavior.

Before:

Accepted (not strict):

cava --versi → treated as --version (prints version)

cava --hel → treated as --help (prints help)

Rejected:

cava --versionx → invalid option

After:

Long options must now match exactly:

--help, --version, --config <path>, --config=<path>

Prefixes/abbreviations are treated as typos and consistently rejected:

cava --versi → cava: error: invalid option

cava --hel → cava: error: invalid option

cava --versionx → cava: error: invalid option

Implementation:

Added pre-validation via validate_cli_long_options() (non-Windows) before the
getopt_long loop to ensure only known long options are accepted, preventing
prefix/abbreviation matching.

Why:

Prevents typos from silently “working”.

Makes CLI behavior consistent and stricter.

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -308,6 +308,55 @@ float *monstercat_filter(float *bars, int number_of_bars, int waves, double mons
     return bars;
 }
 
+#ifndef _WIN32
+static int validate_cli_long_options(int argc, char **argv) {
+    for (int i = 1; i < argc; i++) {
+        const char *arg = argv[i];
+
+        if (arg == NULL) {
+            continue;
+        }
+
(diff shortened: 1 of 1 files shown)


karlstav:
I stand by my policy of "if it's not broken, don't fix it". There are three command line parameters, getopt_long accepts abbreviations for some reason. I can live with it.

PR #742, opened 2026-01-27
feat(build): add WiX installer support for Windows

amnweb:
- Install WiX Toolset and extensions for MSI packaging.
- Create MSI installer for cava with necessary components.
- Upload MSI artifact for distribution.

There is a little difference between this MSI installer and the Inno. I decided not to add a shortcut because you don’t have an icon and probably don’t need, since Cava is a CLI tool, but if you want, we can also include it in the installer.

The license file has been converted from your current one to RTF format. If you need changes, you can modify it.

The MSI installer does not require admin privileges to install because it is installed per user.

The PATH environment variable is added automatically on install and uninstall.

The artifact is included in the workflow build.

What we can do in the future? This MSI is ready for Winget, so you can create a package there once. We can also create a workflow later to update it automatically. The MSI reads the version dynamically, so you don’t need to change anything regarding the version or upgrade code.

https://github.com/user-attachments/assets/49c01174-99a8-4aa0-9e32-d1b7fe2304ef

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -97,7 +97,23 @@ jobs:
       - name: Install dependencies
         run: nuget restore cava_win\cava_win.sln
       - name: Run msbuild
-        run: msbuild /p:Configuration=Debug cava_win\cava_win.sln
+        run: msbuild /p:Configuration=Release cava_win\cava_win.sln
+      - name: Install WiX Toolset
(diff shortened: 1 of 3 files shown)


karlstav:
nice, can you test the installer created from the workflow as well?

https://github.com/karlstav/cava/actions/runs/21405852584/artifacts/5284269732

amnweb:
Tested, everything works fine for me.

PR #746, opened 2026-02-17
chore(md): fix format and markdownlint issues

oxyzenQ:


karlstav:
we don't really use markdown lint here

Issue #594, opened 2024-10-29
Cava is not installing on Arch


Dasa122:
Cava is not building from source or from AUR on arch The arror is in bothe one: (Arch, Terminal: tillix DE:gnome)
/usr/share/automake-1.17/am/depend2.am: error: am__fastdepCC does not appear in AM_CONDITIONAL
/usr/share/automake-1.17/am/depend2.am:   The usual way to define 'am__fastdepCC' is to add 'AC_PROG_CC'
/usr/share/automake-1.17/am/depend2.am:   to 'configure.ac' and run 'aclocal' and 'autoconf' again
/usr/share/automake-1.17/am/depend2.am: error: AMDEP does not appear in AM_CONDITIONAL
/usr/share/automake-1.17/am/depend2.am:   The usual way to define 'AMDEP' is to add one of the compiler tests
/usr/share/automake-1.17/am/depend2.am:     AC_PROG_CC, AC_PROG_CXX, AC_PROG_OBJC, AC_PROG_OBJCXX,
/usr/share/automake-1.17/am/depend2.am:     AM_PROG_AS, AM_PROG_GCJ, AM_PROG_UPC
/usr/share/automake-1.17/am/depend2.am:   to 'configure.ac' and run 'aclocal' and 'autoconf' again

gusruben:
I am having the same issue. Full build logs at <https://hastebin.skyra.pw/urupixosuh.less>

I am running Pipewire 1.2.6 w/ pipewire-pulse.

karlstav:
the real error here is:
```
configure.ac:306: error: _AX_CHECK_GL_MANUAL_LIBS_GENERIC: argument must not be empty
```

the other ones are because aclocal failed but my autogen.sh script still tries to run autoconf which then also fails. (I can look into that later)

whenever one of these things happen in arch it is some new maintainer who has taken over some package and started upgrading it to latest version. In this case [autoconf-archive](https://archlinux.org/packages/extra/any/autoconf-archive/).

there is already a thing for it in your forums:

https://bbs.archlinux.org/viewtopic.php?id=300628

they are suggesting downgrading.

found this in the diff between the two versions:

```
diff autoconf-archive-2023.02.20/m4/ax_check_gl.m4 autoconf-archive-2024.10.16/m4/ax_check_gl.m4 
88c88
< #serial 22
---
> #serial 23
190,191c190
<     AS_IF([test "X$ax_check_gl_manual_libs_generic_extra_libs" = "X"],
[...]

PR #748, opened 2026-02-28
Replace AX_CHECK_GL with manual detection

Yurakaii:
This fixes the issue on arch linux #594, and likely for any distro that is running autoconf-archive >= 2024.10.16 where cava cannot be built with an up-to-date autoconf-archive as there is a broken macro. This replaces the currently broken AC_CHECK_GL macro, with a manual check for opengl. I was initially planning on waiting for an update to autoconf-archive, but it has been over a year, and there is still not a proper fix to this broken macro.


karlstav:
Thanks! The most annoying part is that the bug was actually fixed shortly after it was introduced, but they made a release of the autoconf archive in the mean time and they just don't release that often. So the latest release has now been broken for over a year.

I should have looked at this myself, but the m4 syntax scares me. Brackets, brackets everywhere...

Thanks again!

PR #756, opened 2026-03-20
Move GLSL's current fixed 8-colour gradient system to temporary sampler1D texture generated at runtime.

ninetailedtori:
Working on moving the GLSL shader method itself to generate a temp texture based on the current config structure of setting the gradient_count, then listing the colours as gradient_color_n.
This is then efficiently polled for its 1D coordinate against 0.0–1.0, which then gives us our accurate colour to use in gradient. This works a lot better when using it in shaders as well, but more usefully, allows us to expand the hard-coded 8 colours of gradient to something variable. I welcome any testing on this! Still working on the moving all shaders over, so I can test them all :]


karlstav:
sorry for the extremely late follow up, but I saw the pr was marked as a draft and I assumed it wasn't really ready. The plan sounds legit, I can test it and merge it if it is good to go?

karlstav:
ok tested it now with the default bar spectrum shader, does not work at all it seems. bars are all black.

ninetailedtori:
Yeah sorry I haven't actually completed this one, I ran the shader generation but I don't think I wrote in the actual shader usage to it yet! I'm food-poisoned right now, but feel free to either edit on this PR or if you'd rather, I can write it up once I'm back up, likely tomorrow!

ninetailedtori:
@karlstav finished, test this!

karlstav:
you also added lots of other stuff, touched files and functions not related to the change and added ide type build files. Not sure if this is a symptom of the agentic coding thing going on. I will not even bother testing this until it it stripped back to only the changes related to the topic.

ninetailedtori:
Oh nah this is just my coding hyperfixation leading to me accidentally working on it like it was my own codebase 😭😭😭

I'll strip it back in a bit :3

ninetailedtori:
Alright! I've tried to keep it as minimal here, no changes outside of IDE and formatter hits, outside of the spec of this PR, which does include updating the shaders to use the new gradient texture based on settings (which if disabled, falls back to the same settings that were used initially).

config.c
```diff
@@ -4,7 +4,7 @@
 
 #include <ctype.h>
 #ifndef _WIN32
-#include <iniparser.h>
+#include <iniparser/iniparser.h>
```
karlstav: I guess this " build fix" was committed by accident? Out of curiosity, what distro are you using? This really should not be needed. If the header is under a subfolder that folder should be added as an include flag to the compiler by the build system automatically.

karlstav:
with the bar spectrum shader:

image

master branch:

image

config.c
ninetailedtori: Using arch, header is in /usr/include/iniparser/iniparser.h! could prolly just move it to a specific include yeahhh

ninetailedtori:
Errr, works on mine, is this without gradient enabled? I might've goofed up that shader dammit! Lemme patch that

ninetailedtori:
@karlstav what's your machine? It works fine on mine, I'm going to try to push a fix which might work on yours?

ninetailedtori:
~~Try this one and let me know? It's prolly down to macos metal interp-ing the GL_RGB format wrong, so this is a fallback format to see if it works o-o~~

karlstav:
nope.

i am on intel/nvidia ubuntu 26.04

so the bar_spectrum.frag shader with two colors in the gradient works on your end?

ninetailedtori:
ah, let me test that. one sec.

ninetailedtori:
It was working for me, but it loaded the entire gradient irrespective of the count specified! I've changed the config loading to properly validate colors and load based on count now :3 lemme know if you spot more bugs, I hope not but it is really hot here and I totes am gonna miss smth prolly 😭

ninetailedtori:
Alright, this is...probably as good as it's gonna get for me rn. Gotta take a break for a bit.

Issue #771, opened 2026-05-12
`Audio thread exited unexpectedly.`


doronbehar:
- [x] I went through the [troubleshooting](https://github.com/karlstav/cava/blob/master/README.md#troubleshooting) guide in the readme first.

Are you using cava from a package repository, like AUR?

Yes, [Nixpkgs](https://github.com/NixOS/nixpkgs).

If so, clone the master branch from here, build and run. Is your issue already resolved?

No...

**Describe the bug**

I just run it and I get the message:

```
Audio thread exited unexpectedly.
```

**To Reproduce**

I'm not sure what more steps or instructions should I give...

**Expected behavior**

No error.

**Desktop (please complete the following information):**

 - OS: NixOS, with Gnome
 - Version Nixpkgs' nixos-unstable

**Terminal emulator**
- name: Alacritty
- version: 0.17.0

doronbehar:
Fixed it with:

```ini
[input]
method = pulse
```

But it'd be nice if the error message was nicer of course...

pallaswept:
Unless you really aren't using pipewire in your system, that sounds more like a workaround than a fix. If you check your logs with `journalctl` you will get more information which might be useful so that you can figure out what's wrong so that you can con find a real fix.

doronbehar:
I'm not using pipewire, I'm using pulseaudio indeed. I know my setup is not trivial.

Relevant `journalctl` line printed when I try to launch `cava` with that `method` setting turned off:

```
/nix/store/55sdkgsvgmj0hq2xmrwqjfd5ailzd2hw-xdg-desktop-portal-1.20.4/libexec/xdg-desktop-portal[2630]: Realtime error: Could not get pidns for pid 92266: Could not fstatat ns/pid: Not a directory
```

But I'm surprised `xdg-desktop-portal` is related to `cava`.

pallaswept:
> I'm not using pipewire, I'm using pulseaudio 

In that case you really did fix it :) If you have your system set up like that, t will probably be good practice to double-check configuration for audio apps, because in current day they will most likely assume pipewire is in use.

Though, the[ readme](https://github.com/karlstav/cava#capturing-audio) is a bit confusing here because it says

> This is the default input method if supported on your system.

I'm not so sure that's accurate, as I understand it (I can check the code here if we need a definitive answer), it's the default if cava was built with pipewire support, even if pipewire is not otherwise supported on your system. 

[...]

PR #772, opened 2026-05-13
Additional pipewire error handling

pallaswept:
Add error messages for exits due to stream failures. Avoids an otherwise graceful failure which returns mysterious messages to the user.

These messages can occur when pipewire is in use, so there is no 'maybe try pulse' hint like the other messages which can only occur if the server is missing.

Related: #771

### DIFF
--- a/input/pipewire.c
+++ b/input/pipewire.c
@@ -75,13 +75,13 @@ static void on_stream_state_changed(void *_data, [[maybe_unused]] enum pw_stream
         break;
     case PW_STREAM_STATE_ERROR:
         sprintf(data->cava_audio->error_message,
-            __FILE__ ": Pipewire stream error. Is pipewire running?");
+                __FILE__ ": Pipewire stream error. Is pipewire running?");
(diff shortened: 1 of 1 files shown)


doronbehar:
Works better!

```
Audio thread exited unexpectedly. input/pipewire.c: Pipewire stream not connected.
```

pallaswept:
You weren't joking that would be easy, you had it tested before I could close the tab :D

Thanks for working through that with me. Do you feel like this would be a sufficient solution to the problem you had? Is the error message enough of a hint? I know it's a bit vague, but it kinda needs to be so that it doesn't become misleading if it happens in other scenarios. But I'm open to other wording or anything else you have in mind.

doronbehar:
Yes I think the wording is Not bad. Maybe add at the end: `Is pipewire running?`

pallaswept:
Done! Thank you so much for working through this so we could get it fixed up properly :)

karlstav:
looks great!

PR #786, opened 2026-07-02
Fix iniparser pkg-config name

mohd-akram:


pallaswept:
This is working fine as-is here:

```
BuildRequires:  pkgconfig(iniparser) >= 4.0
```
```
[   25s] checking for libiniparser... no
[   25s] checking for iniparser_load in -liniparser... yes
[   26s] iniparser > 3.2 test OK
```

mohd-akram:
@pallaswept It isn't:

checking for libiniparser... **no**

pallaswept:
> @pallaswept It isn't:
> 
> checking for libiniparser... **no**

I saw the output, but, it is working. There is no description to your PR so it's not clear what your intention is with this change.

karlstav:
it only works because there is a fallback in configure.ac to locate the lib the "old way". using the pkg-config probably never worked.

pallaswept:
> it only works because there is a fallback in configure.ac to locate the lib the "old way". using the pkg-config probably never worked.

I tend to assume that stuff like that is intentional... like, you test for the library by two different methods to catch different names on different distros. If I "fix" something like that then I would break it on some distros.

Obviously by the merge that's not the case here, so yay :)

PR #789, opened 2026-07-25
feat: Implement horizontal_stereo and left_bottom in output config

Tofweod:
This PR adds support for the horizontal_stereo option in example_files/config, which was previously unimplemented.

Based on my understanding (referencing cava's example config), horizontal_stereo renders the left and right channels on the top and bottom halves of the split view respectively. Therefore, mono must be set to average for this to work correctly.

The feature requires:
- orientation set to horizontal
- mono set to average

The left_bottom option defaults to 0. When set to 1, the left channel is placed on the bottom side and the right channel on the top side (i.e., the channels are swapped vertically).

cava_example

### DIFF
--- a/cava.c
+++ b/cava.c
@@ -989,17 +989,28 @@ Keys:\n\
             memset(bars_left, 0, sizeof(float) * number_of_bars / output_channels);
             memset(bars_right, 0, sizeof(float) * number_of_bars / output_channels);
 
-            bars = (int *)malloc(number_of_bars * sizeof(int));
+            if (p.horizontal_stereo) {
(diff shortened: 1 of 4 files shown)


Tofweod:
Sorry for this PR, the horizontal_stereo is implemented by split_stereo, the example config has misled me.

karlstav:
actually the example config is outdated. Forgot to update it after https://github.com/karlstav/cava/commit/99facc54b99309240a1544bce428664c4c314238.

will fix.

Tofweod:
Thx.

